PR #2, opened 2022-02-24
Fixing string display for browser compatibility

rjwills28:
Fixing the issue where strings that are derived from char array are not displayed correctly on Egde and Chrome browsers (see GitHub Issue 3).

### DIFF
--- a/src/types/dtypes.ts
+++ b/src/types/dtypes.ts
@@ -235,7 +235,9 @@ export class DType {
   public static byteArrToString(arr: NumberArray): string {
     let result = "";
     for (let i = 0; i < arr.length; i++) {
-      result += String.fromCharCode(Number(arr[i]));
+      if (Number(arr[i]) !== 0) {
+        result += String.fromCharCode(Number(arr[i]));
+      }
     }
     return result;
   }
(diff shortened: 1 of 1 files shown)


src/types/dtypes.ts
```diff
@@ -235,7 +235,9 @@ export class DType {
   public static byteArrToString(arr: NumberArray): string {
     let result = "";
     for (let i = 0; i < arr.length; i++) {
-      result += String.fromCharCode(Number(arr[i]));
+      if (Number(arr[i]) !== 0) {
+        result += String.fromCharCode(Number(arr[i]));
```
aawdls, line 239: Do you think it makes sense to also else..break here to treat the first `\0` as a terminator? This would save iterating through the whole array when not needed.

aawdls: @Araneidae advises this is a good idea.

Araneidae: Yes, I think it would make a lot of sense to simply treat `'\0'` as a terminator.  Then you can use `strnlen()` to measure the length of the string before processing it.

rjwills28: Yes good idea. I will make that change now and push the update.

aawdls (approved):

PR #3, opened 2022-02-25
Fixing an OPI scaling issue 

rjwills28:
Fixing a scaling issue so that it only scales the opi display when the opi is larger than the screen resolution. If this is not the case then it can scale the OPI height to larger than has already been allocated leading to the bottom being chopped off. This should fix the issue described here: https://github.com/dls-controls/machine-status/issues/2

### DIFF
--- a/src/ui/widgets/Display/display.tsx
+++ b/src/ui/widgets/Display/display.tsx
@@ -25,8 +25,7 @@ const DisplayProps = {
   backgroundColor: ColorPropOpt,
   border: BorderPropOpt,
   macros: MacrosPropOpt,
-  displayHeight: StringPropOpt,
-  displayWidth: StringPropOpt,
+  scaling: StringPropOpt,
   autoZoomToFit: BoolPropOpt
 };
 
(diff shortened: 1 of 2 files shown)


aawdls:
I think we do want it also to scale up when the opi is smaller and I feel it should be possible to accommodate both cases.

Could you scale by  `min(scaleWidthVal, scaleHeightVal)` so that it always fits the biggest dimension at 100%? I haven't tried this out so might be missing something.

aawdls:
Hmmm actually that is essentially the same as what was there before.

Is the problem that the height of a parent element is already set at the smaller value and doesn't adjust when we scale here? Can we overcome that?

rjwills28:
I have updated the logic here for scaling the embedded displays after realising a few things. Firstly, it is now possible to scale the display up or down to make it fill the webpage dimensions (although still preserving the aspect ratio). 
While testing this I realised one issue was that when the height is scaled up (i.e. the display height is made larger than the original height define in the OPI file), the space allocated to the Embedded display does not get updated. So while the opi screen gets enlarged, the space does not and so part of the screen gets cut off. To fix this I had to move the scaling code back into the parent EmbeddedDisplay class (instead of the display class). 
After thorough testing this scaling now performs as expected.

aawdls (approved):

PR #7, opened 2022-03-17
Configure whether a cross to close a display should be shown

rjwills28:
If you load a display from an action button using the 'OPEN_PAGE' action, then the page gets loaded as a dynamicPage which subsequently creates an EmbeddedDisplay. By default this dynamicPage will have a cross in the corner which will close the current EmbeddedDisplay. 
In some cases I do not think this is the required behaviour, i.e. I would expect to open a view and that would be what is displayed on that page. Then if I select another button it would load a different page. I don't think there should be the option to close it as it is a little confusing.
I have added the ability to configure whether this close button is available on a dynamicPage. I have left the default behaviour to be that it is shown (so as to not change the behaviour in other instances). However adding the '`showCloseButton=false`' to the json file will cause the close button not to be shown.

### DIFF
--- a/src/ui/widgets/DynamicPage/dynamicPage.tsx
+++ b/src/ui/widgets/DynamicPage/dynamicPage.tsx
@@ -12,7 +12,12 @@ import { WidgetPropType } from "../widgetProps";
 import { ActionButton } from "../ActionButton/actionButton";
 import { CLOSE_PAGE } from "../widgetActions";
 import { registerWidget } from "../register";
-import { StringProp, InferWidgetProps, BorderPropOpt } from "../propTypes";
+import {
+  StringProp,
+  InferWidgetProps,
+  BorderPropOpt,
+  StringPropOpt
+} from "../propTypes";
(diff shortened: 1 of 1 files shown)


aawdls:
Would you be able to point me at an example so I can visualise the behaviour?

rjwills28:
Absolutely. Attached are some examples I am current working on (ignore the screen formatting). You will see the cross in the upper right corner of the second.
To see it live go to the machine status screens in the cs-web-proto application and select a view from the buttons across the top. Here you will see the screen has a close button in the corner and clicking on that will close that view.

noCross
cross

aawdls (approved):
Thanks. I agree we want to change this.

aawdls:
I expect the difference between when you do and do not want the close button is if opening a new opi replaces the current embedded display or opens a new "tab". For replacing current you probably do not want it. If you are keeping multiple "tabs" then you probably do.

rjwills28:
Yes I agree there is definitely a use case to have the close button. I just thought the inclusion of it should be configurable to cover all use cases and not hardcoded in the library.

PR #10, opened 2022-06-10
Adding the simulated flipflop PV and reloading image.

rjwills28:
This provides a solution to an issue documented in the machine-status repo (https://github.com/dls-controls/machine-status/issues/20) whereby images sourced from a server are not updated after the initial render.
See comments on the this ticket for further details.

### DIFF
--- a/src/connection/sim.ts
+++ b/src/connection/sim.ts
@@ -495,18 +495,11 @@ export class SimulatorPlugin implements Connection {
       updateRate = 100;
       cls = RampPv;
     } else if (nameInfo.protocol.includes("sim://flipflop")) {
-      // Default rate if not configured with flipflop(X)
-      updateRate = 1000;
-      const strCmd = nameInfo.protocol;
-      if (strCmd.includes("(") && strCmd.includes(")")) {
-        const intevalSecStr = strCmd.slice(
-          strCmd.indexOf("(") + 1,
-          strCmd.lastIndexOf(")")
-        );
-        if (intevalSecStr !== "") {
-          updateRate = parseFloat(intevalSecStr) * 1000;
-        }
-      }
+      // Get the time in sec specified between ()
+      const match = nameInfo.protocol.match(/\((.*)\)/);
+      // Default interval if not defined is 1 sec
+      const intervalSecStr = match?.pop() || "1";
+      updateRate = parseFloat(intervalSecStr) * 1000;
       initial = undefined;
       cls = FlipFlopPv;
     } else {
--- a/src/connection/sim.ts
+++ b/src/connection/sim.ts
@@ -318,6 +318,25 @@ class LimitData extends SimPv {
   }
 }
 
+class FlipFlopPv extends SimPv {
+  // Switches between true and false on a loop
+  private value: DType = new DType({ doubleValue: 1 }, DAlarm.NONE, dtimeNow());
+
+  public constructor(...args: SimArgs) {
+    super(...args);
+    setInterval(this.publish.bind(this), this.updateRate);
+  }
+
+  public getValue(): DType {
+    let val = 1;
+    if (this.value.value.doubleValue === 1) {
+      val = 0;
+    }
+    this.value = new DType({ doubleValue: val });
+    return this.value;
+  }
+}
+
 class SimCache {
   private store: { [pvName: string]: SimPv | undefined };
   public constructor() {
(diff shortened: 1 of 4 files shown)


src/connection/sim.ts
```diff
@@ -475,6 +494,21 @@ export class SimulatorPlugin implements Connection {
...
       cls = RampPv;
+    } else if (nameInfo.protocol.includes("sim://flipflop")) {
+      // Default rate if not configured with flipflop(X)
+      updateRate = 1000;
+      const strCmd = nameInfo.protocol;
+      if (strCmd.includes("(") && strCmd.includes(")")) {
+        const intevalSecStr = strCmd.slice(
+          strCmd.indexOf("(") + 1,
+          strCmd.lastIndexOf(")")
+        );
```
aawdls, line 505: I feel a regular expression match here could be more robust. Have you considered this? There could be reasons not to prefer it e.g. speed but I think this is only called on creation.

src/connection/sim.ts
```diff
@@ -475,6 +494,21 @@ export class SimulatorPlugin implements Connection {
...
+        if (intevalSecStr !== "") {
+          updateRate = parseFloat(intevalSecStr) * 1000;
```
aawdls, line 507: Typo in variable name inteval -> interval

src/ui/widgets/Image/image.tsx
```diff
@@ -55,10 +55,16 @@ export const ImageComponent = (
     height: imageHeight
   };
 
+  // Should we be refreshing image instead of using cache
+  let imageFileName = props.imageFile;
+  if (imageFileName.includes("#")) {
```
aawdls, line 60: It looks like we're assuming the `#` will come at the end here - in which case it may be best to make this explicit?

aawdls:
I think this approach is good. 

I feel that the workaround to avoid the caching, by post-processing the image URL, is a bit unsatisfying and it would be nicer to avoid this. But I don't have any better suggestions right now.

A general point on commit messages - can you follow https://confluence.diamond.ac.uk/x/foOQC

aawdls:
Perhaps we can try adding a `no-cache` header on the server side https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control, https://webdock.io/en/docs/webdock-control-panel/optimizing-performance/setting-cache-control-headers-common-content-types-nginx-and-apache

This may avoid having to append the timestamp.

MJGaughran:
I just had a chat with Andy, and I think we agree with the existing pattern of:

1. Using a flipflop to request the image every n seconds
2. Changing a hash / timestamp on the image to prevent cache (so-called 'cache busting' pattern from my brief googling)

We do not want to change the request to include a 'no-cache' directive on the request as this would require that other clients (i.e. Phoebus) are modified to include such a directive too, and it may not always be desirable.

Using no-cache on the server side could always be implemented at a later date when we are more familiar with how to configure something like this.

MJGaughran:
How was this task handled in the original status display application? Did it simply force refresh the image as the display cycled through?

src/ui/widgets/Image/image.tsx
rjwills28: The `#` will come after the .<file_extension> but might not be the last character (e.g. the filename could be image.png#1). We say that we just add `#` to the end of the filename in an OPI if we want it to be refreshed, in which case I could check the last character? Otherwise I could check that it appears after the '.' or use regex but if that is slower we have to keep in mind that this will run on every refresh?

src/connection/sim.ts
rjwills28: Yes this is only called once on the creation of the simulated PV so I think regex would be more robust. I have made this change in the latest commit.

src/connection/sim.ts
rjwills28: Fixed in latest commit.

rjwills28:
I looked at the original status display app and the original internal status webpage to see what they did:
1. Status display app cycles through EDM screens. It uses a 'PNG Image' to display the image from a file location. It contains a property called 'refreshRate' which is set to 30 secs so presumably EDM refreshes the image every X secs automatically.
2. The internal webpage refreshes the whole page ever 10 secs. In the header it contains the line `<meta http-equiv="refresh" content="10; url=index.php">`. I guess this is something we could also do but I'm not sure it's the best idea to refresh the whole page when there's only one element that actually needs to be updated?

aawdls:
Thanks for the info Becky. I think this further supports your approach. You are essentially implementing an analog to case 1.

2 is not desirable. It is definitely preferable to update only the element in question, not reload the whole page. And I think we would have trouble maintaining Phoebus compatibility with 2.

src/ui/widgets/Image/image.tsx
rjwills28: We have decided to keep to logic as is, I.e. does it contain a '#'. A comment has been added to the code to make it clearer why this is.

Issue #11, opened 2022-06-16
Duplicate GraphQL subscriptions sent


rjwills28:
This issue relates to the memory leak found in the Coniql application: https://github.com/dls-controls/coniql/issues/35

While debugging the memory leak in the Coniql application, I came across an issue in the GraphQL subscription mechanism in the cs-web-lib code. I was able to consistently reproduce a case that caused the memory leak in Coniql and one that did not. 

**Memory leak**: This occurs when the a page is loaded which is subscribing to PVs through Coniql, but Coniql has not yet been started (or has been taken down). When the page loads, the `subscribe` commands are sent from the web app. As Coniql has not yet been started, the application performs an `unsubscribe` after a disconnect. When Coniql is started, the web app sends another `subscribe` on reconnection. The previous subscribe on start up has not been cancelled (for some reason) and so 2 subscribes are picked up by Coniql. One of these is old and appears to not be attached to a client - this might be what is causing the memory leak.

**No memory leak**: If Coniql is already running before a web page subscribing to PVs is loaded then there is no memory leak. The `subscribe` is sent once and only once. Note that if you then bring Coniql down and restart it, we will end up in the first situation and get a memory leak. 

Possibly something wrong with the unsubscribe that is not clearing the initial subscription on disconnect.

aawdls:
@AlexanderWells-diamond  We should check whether the same issue exists in the pts app, as likely you followed some similar patterns to cs-web-proto.

AlexanderWells-diamond:
I have certainly seen similar a similar issue - on my development setup I see a GraphQL connection to Coniql established and then closed every minute. Strangely this does not happen in the deployed Production build. I also note that the PTS app does not use the SubscriptionClient, instead using ApolloClient and its `useSubscription` hook. So I suspect its unlikely to be the same issue, but I'll raise one anyway for the weird behaviour in the Development environment.

aawdls:
Fixed with #12

PR #12, opened 2022-06-21
Fix duplicate graphql subscriptions

rjwills28:
This fixes the issue described in https://github.com/dls-controls/cs-web-lib/issues/11. 
Two issues have been addressed: trying to subscribe before the websocket client is connected and the unsubscribe not fully removing the subscription from the websocket (see [comment](https://github.com/dls-controls/cs-web-lib/issues/11#issuecomment-1161558094) in issue for more details).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,7 @@
       "version": "0.1.12",
       "license": "ISC",
       "dependencies": {
+        "apollo-link-retry": "^2.2.16",
         "loglevel": "^1.8.0",
         "redux": "^4.1.2"
       },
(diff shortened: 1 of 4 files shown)


rjwills28:
After taking on board the comments on the current fix, which may introduce a race condition, I did some more research and came up with an alternative solution.

After a lot of digging I am now sure that the Coniql server and underlying tartiflettes subscription code are doing the correct thing - the bug is not here. The issue is on the client side. The underlying problem is in the subscription-transport-ws library, which is used as the SubscriptionClient. Essentially, it does not wait for the acknowledgment receipt from the server when making the connection before clearing the current subscription messages and sending the next. A similar issue has been seen and discussed in this post. The workaround has also been suggested there, which I have implemented and verified (using the network payload mesasges) that it works as it should. 

One final note is that 'subscription-transport-ws' has been deprecated and should be replace with 'graphql-ws'. Hopefully moving to the maintained 'graphql-ws' package will also fix this issue. I am unsure though whether we are in a position yet to start making these updates to the application and so have implemented the workaround for the moment.

PR #15, opened 2023-01-20
Add plot widget

abigailalexander:
I created an XYPlot widget based on the `dawn.xygraph` and `xygraph` widgets that uses plotly.js/react-plotly.js as the plotting library. It is able to mimic most of the existing functionality provided by CSStudio with plotly.js settings, and uses the same default values for properties. It does not currently have the ability to concatenate new data onto already plotted data. 

Some other things to note:

- The names in the .opi file for trace/axis properties are difficult to parse, so I would very much appreciate any feedback on better ways to implement the parsing! 
- Currently cs-web-lib does not correctly resolve nested macro names for PVs, as mentioned in the [Not yet implemented section in the cs-web-proto wiki](https://github.com/dls-controls/cs-web-proto/wiki/Macros). As xygraph widgets contain nested macro names, I have added a small hack to ensure they're parsed correctly in this one use case. Fixing the overall issue seemed like it could be a big divergence from the plot widget focus

### DIFF
--- a/src/types/traces.ts
+++ b/src/types/traces.ts
@@ -54,21 +54,21 @@ export class Trace {
    * csstudio.opibuilder.xygraph.
    */
   public constructor(idx: number) {
-    this.index = idx ?? 0;
-    this.name = this.name ?? "";
-    this.plotMode = this.plotMode ?? 0;
-    this.lineWidth = this.lineWidth ?? 1;
-    this.traceType = this.traceType ?? 0;
-    this.traceColor = this.traceColor ?? TRACE_COLOR_ARRAY[this.index];
-    this.updateDelay = this.updateDelay ?? 100;
-    this.updateMode = this.updateMode ?? 0;
-    this.pointStyle = this.pointStyle ?? 0;
-    this.pointSize = this.pointSize ?? 4;
-    this.concatenateData = this.concatenateData ?? true;
-    this.bufferSize = this.bufferSize ?? 100;
-    this.visible = this.visible ?? true;
-    this.xAxisIndex = this.xAxisIndex ?? 0;
-    this.yAxisIndex = this.yAxisIndex ?? 1;
+    this.index = idx;
+    this.name = "";
+    this.plotMode = 0;
+    this.lineWidth = 1;
+    this.traceType = 0;
+    this.traceColor = TRACE_COLOR_ARRAY[this.index];
+    this.updateDelay = 100;
+    this.updateMode = 0;
+    this.pointStyle = 0;
+    this.pointSize = 4;
+    this.concatenateData = true;
+    this.bufferSize = 100;
+    this.visible = true;
+    this.xAxisIndex = 0;
+    this.yAxisIndex = 1;
   }
 }
 
--- a/src/types/traces.ts
+++ b/src/types/traces.ts
@@ -1,0 +1,77 @@
+import { Color } from "./color";
+
+// TO DO - find out which variables always appear and which don't and add ?
+
+export class Trace {
+  public name?: string;
+  public plotMode?: number; //convert to value instead of num?
+  public lineWidth?: number;
+  public traceType?: number; //convert?
+  public traceColor?: Color;
+  public updateDelay?: number;
+  public updateMode?: number; //convert
+  public pointStyle?: number; //convert
+  public pointSize?: number;
+  public concatenateData?: boolean;
+  public bufferSize?: number;
+  public visible?: boolean;
+  public antiAlias?: boolean;
+  public xPv?: string;
+  public xPvValue?: number;
+  public xAxisIndex?: number;
+  public yPv?: string;
+  public yPvValue?: number;
+  public yAxisIndex?: number;
+
+  public constructor(
+    name: string,
+    plotMode: number,
+    lineWidth: number,
+    traceType: number,
+    traceColor: Color,
+    updateDelay: number,
+    updateMode: number,
+    pointStyle: number,
+    pointSize: number,
+    concatenateData: boolean,
+    bufferSize: number,
+    visible: boolean,
+    antiAlias: boolean,
+    xPv: string,
+    xPvValue: number,
+    xAxisIndex: number,
+    yPv: string,
+    yPvValue: number,
+    yAxisIndex: number
+  ) {
+    this.name = name;
+    this.plotMode = plotMode;
+    this.lineWidth = lineWidth;
+    this.traceType = traceType;
+    this.traceColor = traceColor;
+    this.updateDelay = updateDelay;
+    this.updateMode = updateMode;
+    this.pointStyle = pointStyle;
+    this.pointSize = pointSize;
+    this.concatenateData = concatenateData;
+    this.bufferSize = bufferSize;
+    this.visible = visible;
+    this.antiAlias = antiAlias;
+    this.xPv = xPv;
+    this.xPvValue = xPvValue;
+    this.xAxisIndex = xAxisIndex;
+    this.yPv = yPv;
+    this.yPvValue = yPvValue;
+    this.yAxisIndex = yAxisIndex;
+  }
+}
+
+export class Traces {
+  public count: number;
+  public traceOptions: Trace[];
+
+  public constructor(count: number, traces: Trace[]) {
+    this.count = count;
+    this.traceOptions = traces;
+  }
+}
--- a/src/types/traces.ts
+++ b/src/types/traces.ts
@@ -1,68 +1,74 @@
 import { Color } from "./color";
 
-// TO DO - find out which variables always appear and which don't and add ?
+// Colour chart for traces
+
+const TRACE_COLOR_ARRAY: { [key: number]: Color } = {
+  0: new Color("rgb(21, 21, 196)"), // blue
+  1: new Color("rgb(242, 26, 26)"), // red
+  2: new Color("rgb(33, 179, 33)"), // green
+  3: new Color("rgb(0, 0, 0)"), // black
+  4: new Color("rgb(128, 0, 255)"), // violet
+  5: new Color("rgb(255, 170, 0)"), // (darkish) yellow
+  6: new Color("rgb(255, 0, 240)"), // pink
+  7: new Color("rgb(243, 132, 132)"), // peachy
+  8: new Color("rgb(0, 255, 11)"), // neon green
+  9: new Color("rgb(0, 214, 255)"), // neon blue
+  10: new Color("rgb(114, 40, 3)"), // brown
+  11: new Color("rgb(219, 128, 4)"), // orange
+  12: new Color("rgb(21, 21, 196)"), // Repeat colours until 20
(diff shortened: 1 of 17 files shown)


src/ui/widgets/XYPlot/xyPlotOptions.ts
```diff
@@ -0,0 +1,259 @@
...
+      // Use the last n points
+      tmpY = tmpY.slice(-numPoints);
+      tmpX = tmpX.slice(-numPoints);
+    } else if (numPoints < tmpY.length) {
+      // Use the first n points
+      tmpY = tmpY.slice(0, numPoints);
+      tmpX = tmpX.slice(0, numPoints);
+    }
+
+    data.x = tmpX;
```
aawdls, line 120: Truncating will break your concat indexing above - as you probably know

src/ui/widgets/XYPlot/xyPlotOptions.ts
```diff
@@ -0,0 +1,259 @@
...
+  newAxis: NewAxisSettings,
+  dataSet: any[]
+): NewAxisSettings {
+  // If autoscale not enabled, do not rescale axes
+  if (!oldAxis.autoScale) return newAxis;
+  // Find axis limits
+  const [axMin, axMax]: [number, number] = newAxis.range;
+  // Determine if we are using x or y data
+  let dataType = "x";
+  if (oldAxis.index > 0 && oldAxis.yAxis) dataType = "y";
```
aawdls, line 211: Bit of redundancy here as you have numbered them such that this is always so

src/ui/widgets/XYPlot/xyPlotOptions.ts
```diff
@@ -0,0 +1,259 @@
...
+  // Combine all data we need into an array of arrays
+  const allData = dataSet.map(val => val[dataType]);
```
aawdls, line 213: This might imply some copying which I would be wary of on performance grounds. There may be a more efficient way.

aawdls: This is where I don't know how TS handles references; in python it wouldn't copy but here I am not sure.

You could iterate and take the max of the maxes and the min of the mins of each array.

src/ui/widgets/XYPlot/xyPlotOptions.ts
```diff
@@ -0,0 +1,259 @@
...
+
+/**
+ * Round a decimal value up or down to the nearest
+ * significant figure.
+ * @param value decimal value to round
+ * @param mag magnitude to round to
+ * @param roundType 0 to round down, 1 to round up
+ * @returns rounded number
+ */
+export function roundValue(value: number, roundType: number): number {
```
aawdls, line 241: Pretty surprised you had to implement this yourself.

aawdls: Still, probably better to do this than add yet another dependency.

aawdls (commented):
Thanks. Only minor suggestions.

src/ui/widgets/EmbeddedDisplay/opiParser.ts
```diff
@@ -454,6 +458,146 @@ function opiParseLabelPosition(props: any): string {
...
+        } else if (COLOR_PROPS.includes(newName)) {
+          parsedProps[newName] = opiParseColor(props[item]);
+        } else if (STRING_PROPS.includes(newName)) {
+          parsedProps[newName] = opiParseString(props[item]);
+        } else if (FONT_PROPS.includes(newName)) {
+          parsedProps[newName] = opiParseFont(props[item]);
+        }
+      }
+    } catch {
+      // do nothing
```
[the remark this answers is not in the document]
abigailalexander: Sometimes properties exist in an opi file but don't contain any value e.g. `<axis_0_axis_title />`, so when you try to pass them to a function like opiParseString it throws an exception

src/ui/widgets/XYPlot/xyPlotOptions.ts
abigailalexander: The TS `.floor()` and `.ceil()` methods always round to integers. While there are methods that can round decimals to a specified significant figure after the dot (e.g. `toFixed()`), there isn't a way to specify whether it should round up or down. I was also surprised that something like this doesn't seem to exist already in TS...

src/types/traces.ts
```diff
@@ -0,0 +1,90 @@
...
+  10: new Color("rgb(114, 40, 3)"), // brown
+  11: new Color("rgb(219, 128, 4)"), // orange
+  12: new Color("rgb(21, 21, 196)"), // Repeat colours until 20
+  13: new Color("rgb(242, 26, 26)"),
+  14: new Color("rgb(33, 179, 33)"),
+  15: new Color("rgb(0, 0, 0)"),
+  16: new Color("rgb(128, 0, 255)"),
+  17: new Color("rgb(255, 170, 0)"),
+  18: new Color("rgb(255, 0, 240)"),
+  19: new Color("rgb(243, 132, 132)")
```
rjwills28, line 25: Does this mean that we will limit the number of traces on a plot to 20? I see that CS-Studio also limits to 20 traces. What would happen if there were more - do they just get ignored?

src/ui/widgets/EmbeddedDisplay/opiParser.ts
```diff
@@ -454,6 +458,146 @@ function opiParseLabelPosition(props: any): string {
...
+
+/**
+ * Converts property name from snake case to camel
+ * case. This is needed to dynamically sort through
+ * all of each traces properties, although it's a
+ * bit messy and clunky.
+ * @param propName string to convert
+ * @returns newly formatted string
+ */
+function snakeCaseToCamelCase(propName: string) {
```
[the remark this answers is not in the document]
rjwills28: It seems we don't really have a 'utils' file for tings like this. Although is the use of this function still specific to parsing? I'm not sure we would need to use it anywhere else. In that case maybe it could be moved to parser.ts?

src/ui/widgets/XYPlot/xyPlot.tsx
```diff
@@ -0,0 +1,129 @@
...
+    for (let i = 0; i < len; i++) {
+      plotLayout[axisNames[i]] = newAxisOptions.shift();
+    }
+    return (
+      <div className={"showBorder"} style={style}>
+        <Plot data={dataSet} layout={plotLayout} />
+      </div>
+    );
+  }
+  return <></>;
```
rjwills28, line 117: If the checks fail then it currently just displays an empty box. It's a little unclear at this point where things have gone wrong. It might be good to display a message in the box saying that something is missing from the opi file. This is done in other places when the opi file can't be found for example: https://github.com/dls-controls/cs-web-lib/blob/3453939456917a6e2a1fbafed9042dc9045ebfa6/src/ui/widgets/EmbeddedDisplay/embeddedDisplay.tsx#L94

src/ui/widgets/XYPlot/xyPlot.tsx
```diff
@@ -0,0 +1,129 @@
...
+    titleFont,
+    showLegend = true,
+    showPlotBorder,
+    // showToolbar, // TO DO - do we want a toolbar as well?
+    traces,
+    axes
+  } = props;
+  // TO DO - having all these checks is not ideal
+  if (
+    value?.value.arrayValue &&
```
rjwills28, line 56: Will we only support waveform PVs? In CS-Studio the xyplot also plots 'live' data as the PV value changes? I guess this will need the concatenate data ability which I know is not working yet.

src/ui/widgets/XYPlot/xyPlot.tsx
```diff
@@ -0,0 +1,129 @@
...
+    const bytesPerElement = value.value.arrayValue.BYTES_PER_ELEMENT;
+    // If data exists, creates traces to plot
+    const dataSet = createTraces(traces, value, bytesPerElement);
+    // Set up style
+    let style: CSSProperties = {};
+    if (showPlotBorder) {
+      style = { border: "1px solid black", padding: "1px" };
+    }
+    const font = titleFont?.css();
+    // Sometimes font is a string with "px" on the end....
```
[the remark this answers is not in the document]
rjwills28: We do have to do this in other place. For example. the height and width are also sent with "px": https://github.com/dls-controls/cs-web-lib/blob/3453939456917a6e2a1fbafed9042dc9045ebfa6/src/ui/widgets/EmbeddedDisplay/embeddedDisplay.tsx#L65
So it might be worth turning this into a standalone function that can be re-used?

src/ui/widgets/XYPlot/xyPlotOptions.test.ts
```diff
@@ -0,0 +1,514 @@
...
+  roundValue
+} from "./xyPlotOptions";
+
+const DATA = [0, 3, 4, 8];
+const ARRAY_DATA = new DType({
+  arrayValue: new Float32Array(DATA)
+});
+
+describe("round values", (): void => {
+  test("returns zero if values zero", (): void => {
```
rjwills28, line 19: I was just curious as to why we sometimes use `test()` and other times use `it()` to declare a test. The Jest documentation says that `it()` is an alias for `test()` so functionally they are the same. Some users suggest that `it()` makes the tests more readable. Should we look at picking one and remaining consistent throughout the code?

src/ui/widgets/XYPlot/xyPlotOptions.ts
rjwills28: I think `.map()` performs a shallow copy of the object. Iterating may indeed be quicker.

src/ui/widgets/XYPlot/xyPlotOptions.ts
rjwills28: A useful function - perhaps it could be moved to a common location? Maybe we should create a new 'utils' file with some of these methods in?

rjwills28 (commented):
This is looking really good. I've also built the library and tested the xyplot which behaves as expected. We benefit from some nice features of the plotly lib allowing you to zoom and move around the plot.
I've just made a few minor suggestions to look over and discuss. Thanks.

src/types/traces.ts
abigailalexander: I based this off CS-Studio's limit of 20, although I don't know if it's something we want to continue to enforce or allow for more traces? You raise a good point I hadn't considered though, I don't know what CSS did if too many traces were provided. In this case, I think having more than 20 traces would throw an index out of range error when it attempts to allocate a colour, which it definitely shouldn't do

src/ui/widgets/XYPlot/xyPlot.tsx
abigailalexander: We would want to support live data plots too, which I assume would be plotting the live data against time as it comes in? You're correct that it would use the concatenate data, and also reconfiguring to have time as an axis instead of index. I believe the plan is to add in support for time series plots (and also multiple PVs for a single widget), but I haven't got around to it quite yet....

PR #16, opened 2023-01-31
Webcam widget

abigailalexander:
A webcam widget that displays MJPEG streams (or other compatible formats) in a HTML image. The widget expands to fill the entire window and maintains aspect ratio when window is resized in any direction. If a connection cannot be made, the alt text displays an error message informing the viewer of this.

### DIFF
--- a/src/ui/widgets/Webcam/webcam.module.css
+++ b/src/ui/widgets/Webcam/webcam.module.css
@@ -1,0 +1,5 @@
+img {
+    width: 100%;
+    height: 100%;
+    object-fit: contain;
+}
--- a/src/ui/widgets/Webcam/webcam.test.tsx
+++ b/src/ui/widgets/Webcam/webcam.test.tsx
@@ -34,7 +34,7 @@ describe("<Webcam />", (): void => {
     );
     expect(
       getByAltText(
-        "Connection could not be made to Webcam MJPEG stream at http://fake-stream.diamond.ac.uk/video.mjpg"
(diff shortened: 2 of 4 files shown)


aawdls:
I don't have any feedback on the code itself.

Two thoughts I had about the widget.

- It's called "webcam" and has "webcam" in the strings for the alt text etc. Can we think of any other streams that we'd use this for that are not webcams? I can't at the moment. We could change this later if needed.
- Is there any point trying to relate it to the existing image widget. I think not, as it is a distinct purpose and combining would make it more complicated rather than less.

abigailalexander:
I can't think of any other types of streams that aren't webcams, at least for the current custom SDMs. I also think relating it to the current image widget would just confuse things as they have different size and alt text requirements.

One thing that I have just noticed: I was treating Webcam as a component and not a widget initially, and created it mimicking that style. Now that it is in the widgets directory, would it be worth updating the code to match the widget conventions? It would make very little functional difference but ensure that it is consistent with the other widgets...

aawdls:
> I can't think of any other types of streams that aren't webcams, at least for the current custom SDMs. I also think relating it to the current image widget would just confuse things as they have different size and alt text requirements.

Agreed

> One thing that I have just noticed: I was treating Webcam as a component and not a widget initially, and created it mimicking that style. Now that it is in the widgets directory, would it be worth updating the code to match the widget conventions? It would make very little functional difference but ensure that it is consistent with the other widgets...

I think that is a good idea

Issue #22, opened 2023-03-03
Console errors in tests


rjwills28:
Something is missing in the test environment causing multiple console.errors to be printed when running the tests. The error is related to the use of the plotly.js module and the fact that the tests do not have a canvas to paint on (I think? - see error below.). Note that these do not cause the tests to fail but it is a bit annoying to have to see all these errors printed. You can also see them in the recent CI jobs.

```
console.error
      Error: Not implemented: HTMLCanvasElement.prototype.getContext (without installing the canvas npm package)
          at module.exports (/home/cnuser/cs-web-proto/cs-web-lib/node_modules/jsdom/lib/jsdom/browser/not-implemented.js:9:17)
          at HTMLCanvasElementImpl.getContext (/home/cnuser/cs-web-proto/cs-web-lib/node_modules/jsdom/lib/jsdom/living/nodes/HTMLCanvasElement-impl.js:42:5)
          at HTMLCanvasElement.getContext (/home/cnuser/cs-web-proto/cs-web-lib/node_modules/jsdom/lib/jsdom/living/generated/HTMLCanvasElement.js:131:58)
          at Object.42505 (/home/cnuser/cs-web-proto/cs-web-lib/node_modules/plotly.js/dist/plotly.js:156927:42)
          at __webpack_require__ (/home/cnuser/cs-web-proto/cs-web-lib/node_modules/plotly.js/dist/plotly.js:222552:42)
          at Object.26787 (/home/cnuser/cs-web-proto/cs-web-lib/node_modules/plotly.js/dist/plotly.js:98289:12)
          at __webpack_require__ (/home/cnuser/cs-web-proto/cs-web-lib/node_modules/plotly.js/dist/plotly.js:222552:42)
          at Object.68868 (/home/cnuser/cs-web-proto/cs-web-lib/node_modules/plotly.js/dist/plotly.js:98275:14)
          at __webpack_require__ (/home/cnuser/cs-web-proto/cs-web-lib/node_modules/plotly.js/dist/plotly.js:222552:42)
          at Object.21641 (/home/cnuser/cs-web-proto/cs-web-lib/node_modules/plotly.js/dist/plotly.js:481:18) undefined
```

rjwills28:
Fixed in PR #23 .

PR #23, opened 2023-03-03
Fix test console errors

rjwills28:
A lot of console errors get printed when running the tests. These do not stop the test from failing but are quite annoying. See full details in issue #22.

The fix is to import the "jest-canvas-mock" module in the tests that end up calling the plotly.js module down the line.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -35,6 +35,7 @@
         "eslint-plugin-import": "^2.20.0",
         "eslint-plugin-prettier": "^3.1.2",
         "eslint-plugin-unicorn": "^33.0.0",
+        "jest-canvas-mock": "^2.4.0",
         "loglevel": "^1.8.0",
         "prettier": "^2.2.1",
         "react-id-generator": "^3.0.0",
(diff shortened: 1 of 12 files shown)


AlexanderWells-diamond:
So I'll preface this by saying I know very little about Javascript testing or Jest.

It seems wrong that you have to manually import this into each and every test file that needs it. Especially as it'll be easy to forget to do this in the future - missing it only produces warnings, not errors, which could easily be overlooked when writing new tests. 

I had a look through the docs and think a [Setup](https://www.npmjs.com/package/jest-canvas-mock#Setup) section in `package.json` (or a separate Setup File, documented further down that page) may allow you to automatically bring this dependency into all tests without having to modify each one?

I suppose this approach is only valid if importing this module has no side-effects on tests that don't need it.

rjwills28:
Thanks for the suggestion! It is indeed possible to require the module for all tests and it has no impact on other tests that don't need it. Much cleaner solution.

aawdls (approved):

PR #26, opened 2023-03-24
Force react-plotly.js to use plotly.js-basic-dist

abigailalexander:
This fixes an issue that was observed in applications using cs-web-lib (e.g. machine-status) and caused by the large size of react-plotly.js. The library and Plot component are so large that the size of cs-web-lib also increases. This results in operations like `npm run lint` and `npm run build` failing in applications that import cs-web-lib due to running out of memory. Increasing the available memory using `NODE_OPTIONS=--max-old-space-size=XXXX` works, but doesn't address the core issue.

I have implemented a fix for this that is referenced in [this issue](https://github.com/plotly/react-plotly.js/issues/135#issuecomment-501398125). It involves swapping the plotly.js library that react-plotly.js uses for a stripped down version plotly.js-basic-dist which only contains scatter, pie and bar charts. This decreases the package size and doesn't affect functionality as we only need the scatter plot.

To check that this branch fixes the issue:
```cd cs-web-lib
git pull
git checkout modify-plotly-import
npm install --save-dev
npm run rollup
npm link
cd ../machine-status (or other application)
npm uninstall @dls-controls/cs-web-lib
npm link @dls-controls/cs-web-lib
npm install path/to/cs-web-lib
npm run lint
npm run build
```

This creates a local build of cs-web-lib and installs it as a package into the application. When testing it myself, the memory issues were not present with this branch. I also checked the appearance of the plot itself briefly in cs-web-proto by implementing these same changes and it appears to look and behave in an identical way.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,6 +12,7 @@
         "apollo-link-retry": "^2.2.16",
         "bufferutil": "^4.0.7",
         "loglevel": "^1.8.0",
+        "plotly.js": "^2.20.0",
         "plotly.js-basic-dist": "^2.20.0",
         "react-plotly.js": "^2.6.0",
         "redux": "^4.1.2",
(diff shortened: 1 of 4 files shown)


rjwills28 (approved):
This looks good. Following the instructions I am now able to run lint, build and start for the machine-status app (which I also couldn't before). 
I also tested the graph widget with a simple opi and this displayed as expected so it looks like functionality is the same.

PR #27, opened 2023-04-03
Update webcam to display multiple on page

abigailalexander:
The previous implementation of the webcam widget only allowed for a single one to be displayed per page - this was due to auto-sizing of the widget being based on the window size and each webcam widget having the same id. I have updated this now so that the scaling of webcams depends on the size of parent element, and each accepts a name/id to differentiate them. The tests have also been updated for the addition of an id/name.

The webcam sizing isn't ideal as if you attempt to make the page smaller than the original webcam size l it does not shrink, but instead starts to crop the edges. I haven't figured out the exact CSS needed to achieve this.

### DIFF
--- a/src/ui/widgets/Webcam/webcam.tsx
+++ b/src/ui/widgets/Webcam/webcam.tsx
@@ -47,7 +47,7 @@ export const WebcamComponent = (
 function onError(event: any) {
   const image = document.getElementById(event.target.id) as HTMLImageElement;
   const alt = `Connection to webcam at ${image.src} failed`;
-  (document.getElementById(event.target.id) as HTMLImageElement).alt = alt;
+  image.alt = alt;
 }
 
 /**
@@ -56,7 +56,7 @@ function onError(event: any) {
 function onLoad(event: any) {
   const image = document.getElementById(event.target.id) as HTMLImageElement;
   const alt = `Webcam MJPEG stream at ${image.src}`;
-  (document.getElementById(event.target.id) as HTMLImageElement).alt = alt;
+  image.alt = alt;
 }
 
 const WebcamWidgetProps = {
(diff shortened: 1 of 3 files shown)


aawdls:
>The webcam sizing isn't ideal as if you attempt to make the page smaller than the original webcam size l it does not shrink, but instead starts to crop the edges. I haven't figured out the exact CSS needed to achieve this.

This is not currently a requirement for us; these widgets are not trying to be responsive to window size

src/ui/widgets/Webcam/webcam.tsx
```diff
@@ -18,20 +19,17 @@ const WebcamProps = {
...
-        height: height,
-        width: width,
         fontSize: "14px", // Set properties for image alt text
         fontWeight: "bold",
         color: "#a6190f"
       }}
     >
       <img
-        id="stream"
+        id={props.name}
```
aawdls, line 32: Is the use of name and id consistent with other widgets?

src/ui/widgets/Webcam/webcam.tsx
```diff
@@ -46,44 +44,19 @@ export const WebcamComponent = (
  * Returns an error visible to the user when the MJPEG stream
  * at the given URL cannot be reached.
  */
-function onError() {
-  const image = document.getElementById("stream") as HTMLImageElement;
+function onError(event: any) {
+  const image = document.getElementById(event.target.id) as HTMLImageElement;
   const alt = `Connection to webcam at ${image.src} failed`;
-  (document.getElementById("stream") as HTMLImageElement).alt = alt;
+  (document.getElementById(event.target.id) as HTMLImageElement).alt = alt;
```
aawdls, line 50: could use `image.alt`  since you have defined `image`?

aawdls (approved):

src/ui/widgets/Webcam/webcam.tsx
abigailalexander: Quite a few other widgets use the name prop e.g. GroupBox, usually in conjuction with the name attribute in the HTML element. The name attribute doesn't exist on the image element however, so I'm using id instead.

Issue #28, opened 2023-04-28
Scaling issue when an embeddedDisplay does not take up full display  


rjwills28:
When trying to build a view that includes multiple embedded display or an embedded display and another elements (for example a webcam) the auto-scaling of the embeddedDisplay does not work correctly. This is because an Absolute position is used to define the location of the embeddedDisplay and this does not take up the whole of the browser display window. It means that you end up cutting off parts of the embeddedDisplay - example below. The amount visible depends on the size of the browser window hence suggestion that the scaling is not working properly.

![size1](https://user-images.githubusercontent.com/45264327/235144057-d14e3be2-30b7-4b94-8d28-448dd2cc6827.png)

I think the issue is that the previous scaling we implemented was based on the browser window size and does not take into account this case where there is a position allocated for the embeddedDisplay. This needs fixing the cs-web-lib.

rjwills28:
Also noticed some bad scaling issues when refactoring cs-web-proto to use cs-web-lib. These OPIs look OK in the original so a change in cs-web-lib has caused the scaling to got awry in a few cases.

![Screenshot from 2023-05-15 17-33-02](https://github.com/dls-controls/cs-web-lib/assets/45264327/37827a76-7c77-41e9-b263-b55677c6671e)

PR #34, opened 2023-05-19
Fix autoZoomToFit scaling for all observed use-cases

rjwills28:
The autoscaling to fit the window size does get a little complicated as there multiple use cases of this. The problem relates to what you define as the scale origin. This differs for different configurations. This PR fixes/maintains behaviour for the following cases:
- machine-status displays - these displays are central in the window and so their scaling origin is "center top". This will be the default for embedded displays unless specified.
- webcam displays - the original scaling issue referenced in #28 would be fixed with this re-work, but note it was already fixed in the machine-status app by changing the positions to be absolute instead of a percentage of the window.
- dynamicPage - layouts in cs-web-proto use dynamic pages to load in an OPI. These need to have a scaling origin of "0 0" to expand/contract into their designated area within the window.
- Linking Containers - OPIs which embed other OPIs are parsed as embeddedDisplays with embedded embeddedDisplays. We have to be careful here as if both the parent and child OPI has 'autoZoomToFit' as true then the scaling gets applied twice, once to the parent container and again to the child. I have created a special override parameter that can stop scaling of the child if the parent has already been scaled. Equally if the parent is not scaled but the child should be then we still need to expand the space in the parent container otherwise some of the child may be cut off when expanded.

I have tested these scaling fixes against all displays in cs-web-proto, machine-status display screens, machine-status web view and machine-status sdms. Behaviour is maintained in all machine-status screens and improved for cs-web-proto.

### DIFF
--- a/src/ui/widgets/Display/display.tsx
+++ b/src/ui/widgets/Display/display.tsx
@@ -26,7 +26,8 @@ const DisplayProps = {
   border: BorderPropOpt,
   macros: MacrosPropOpt,
   scaling: StringPropOpt,
-  autoZoomToFit: BoolPropOpt
+  autoZoomToFit: BoolPropOpt,
+  scalingOrigin: StringPropOpt
 };
 
 // Generic display widget to put other things inside
@@ -54,11 +55,15 @@ export const DisplayComponent = (
   style["position"] = "relative";
   style["overflow"] = props.overflow;
   style["height"] = "100%";
-  if (props.autoZoomToFit) {
-    const internalScale =
-      typeof props.scaling === "undefined" ? "1" : props.scaling;
+  const internalScale =
+    typeof props.scaling === "undefined" ? "1" : props.scaling;
+  if (props.autoZoomToFit && internalScale !== "1") {
     style["transform"] = "scale(" + internalScale + ")";
-    style["transformOrigin"] = "center top";
+    const scalingOrigin =
+      typeof props.scalingOrigin === "undefined"
+        ? "center top"
+        : props.scalingOrigin;
+    style["transformOrigin"] = scalingOrigin;
   }
   return (
     <MacroContext.Provider value={displayMacroContext}>
--- a/src/ui/widgets/Display/display.tsx
+++ b/src/ui/widgets/Display/display.tsx
@@ -18,6 +18,7 @@ import {
   MacroContext,
   MacroContextType
 } from "../../../types/macros";
+import { getOptionalValue } from "../utils";
 
 const DisplayProps = {
   children: ChildrenPropOpt,
@@ -55,14 +56,15 @@ export const DisplayComponent = (
   style["position"] = "relative";
   style["overflow"] = props.overflow;
   style["height"] = "100%";
-  const internalScale =
-    typeof props.scaling === "undefined" ? "1" : props.scaling;
+  // Check whether the scaling property has been set and if
+  // not set the scaling to 1 (i.e. no scaling)
+  const internalScale = getOptionalValue(props.scaling, "1");
+  // Only configure the scaling transform if autoZoomToFit is
+  // true AND the scale factor provided is not 1 where 1
+  // implies no scaling
   if (props.autoZoomToFit && internalScale !== "1") {
     style["transform"] = "scale(" + internalScale + ")";
-    const scalingOrigin =
-      typeof props.scalingOrigin === "undefined"
-        ? "center top"
-        : props.scalingOrigin;
+    const scalingOrigin = getOptionalValue(props.scalingOrigin, "center top");
     style["transformOrigin"] = scalingOrigin;
   }
   return (
--- a/src/ui/widgets/EmbeddedDisplay/embeddedDisplay.tsx
+++ b/src/ui/widgets/EmbeddedDisplay/embeddedDisplay.tsx
@@ -22,7 +22,7 @@ import {
 import { GroupBoxComponent } from "../GroupBox/groupBox";
 import { useOpiFile } from "./useOpiFile";
 import { useId } from "react-id-generator";
-import { trimFromString } from "../utils";
+import { getOptionalValue, trimFromString } from "../utils";
 
 const EmbeddedDisplayProps = {
   ...WidgetPropType,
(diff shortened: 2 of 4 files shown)


src/ui/widgets/Display/display.tsx
```diff
@@ -54,11 +55,15 @@ export const DisplayComponent = (
   style["position"] = "relative";
   style["overflow"] = props.overflow;
   style["height"] = "100%";
-  if (props.autoZoomToFit) {
-    const internalScale =
-      typeof props.scaling === "undefined" ? "1" : props.scaling;
+  const internalScale =
```
aawdls, line 58: Some comments here explaining this logic would be good. I know you have written some further down but if you come to this  in isolation it is hard to understand

src/ui/widgets/EmbeddedDisplay/embeddedDisplay.tsx
```diff
@@ -40,41 +41,56 @@ export const EmbeddedDisplay = (
...
+    typeof props.autoZoomToFitOverride === "undefined"
+      ? true
+      : props.autoZoomToFitOverride;
 
-  // Apply the height to the top level if relative layout and none have been provided
-  if (props.position instanceof RelativePosition) {
-    props.position.height = props.position.height || description.height;
-    props.position.width = props.position.width || description.width;
-  }
+  // Get the screen height
```
aawdls, line 50: Is there the possibility to factor any of this stuff out into a function? This may be complicated as you are operating on props etc

src/ui/widgets/EmbeddedDisplay/embeddedDisplay.tsx
```diff
@@ -40,41 +41,56 @@ export const EmbeddedDisplay = (
...
+  const heightString =
+    typeof description.position.height === "undefined"
+      ? "10"
```
aawdls, line 53: Why the number 10 in particular here?

src/ui/widgets/EmbeddedDisplay/embeddedDisplay.tsx
```diff
@@ -40,41 +41,56 @@ export const EmbeddedDisplay = (
 
   log.debug(description);
 
-  // Apply the height to the top level if relative layout and none have been provided
-  if (props.position instanceof RelativePosition) {
-    props.position.height = props.position.height || description.height;
-    props.position.width = props.position.width || description.width;
-  }
+  // Check whether to override OPI autoZoomToFit
```
aawdls, line 44: Please elaborate in this comment why/how that would be done as well - and can this property be named so as to make it clearer what a value of true or false would do? i.e. we are overriding it but in which sense?

src/ui/widgets/EmbeddedDisplay/embeddedDisplay.tsx
```diff
@@ -40,41 +41,56 @@ export const EmbeddedDisplay = (
...
+      if (description.autoZoomToFit && autoZoomToFitOverride) {
+        // If we have scaled the parent then do not scale
+        // the child display as this will result in double scaling
+        displayChildern[i].autoZoomToFitOverride = false;
+      } else {
+        // Otherwise pass on the scalingOrigin so if the child should
+        // be scaled, then it will use the same transform-origin
+        displayChildern[i].scalingOrigin = props.scalingOrigin;
+        // Update the parent container area to be the full area for the child
+        // embedded display so it can be scale into it if needed
```
aawdls, line 86: scaled

src/ui/widgets/EmbeddedDisplay/embeddedDisplay.tsx
```diff
@@ -40,41 +41,56 @@ export const EmbeddedDisplay = (
...
+    typeof description.position.height === "undefined"
+      ? "10"
+      : description.position.height;
+  const widthString =
+    typeof description.position.width === "undefined"
+      ? "10"
+      : description.position.width;
 
+  // Height and width property will take form "<num>px" so trim
+  const heightInt = trimFromString(heightString);
```
aawdls, line 61: Would anyone have used other units, e.g. % or pt?

aawdls (changes requested):
Thanks for this. I have made some suggestions, mainly towards making it a bit easier to understand on first reading.

Just to confirm the props we are setting are in addition to what comes from the widgets in the opi file and we are not deviating from the CSS behaviour?

src/ui/widgets/EmbeddedDisplay/embeddedDisplay.tsx
abigailalexander: Also, it seems like there's a typo in the const name used here, `displayChildern` instead of `displayChildren`?

src/ui/widgets/Display/display.tsx
rjwills28: I have now added a comment to clarify this how this value is determined.

src/ui/widgets/EmbeddedDisplay/embeddedDisplay.tsx
rjwills28: Yes, very good idea as I use this method of determining the value of an optional parameter multiple times. I have created a function and moved it to 'utils.ts'. I then use this in both the embeddedDisplay and display code.

src/ui/widgets/EmbeddedDisplay/embeddedDisplay.tsx
rjwills28: This was random. The height is always defined in the opi file so this shouldn't be a problem, but I have updated it so that the default should it not be set is the window height/width.

src/ui/widgets/EmbeddedDisplay/embeddedDisplay.tsx
rjwills28: Fixed both spelling mistakes.

rjwills28:
Thanks both for the reviews.

@aawdls to clarify, the new props I have added are in addition to those that come from the OPI file. We are not deviating from the behaviour of CSS. In CSS, if both the parent opi and its contained linking container opi has autoZoomToFit=true then the end result is a scaled display to the size of the window. Our original implementation in cs-web-lib did not do this as it applied the scale factor twice, once to the child and then again to the parent so it did not zoom to fit the window size correctly. The new props just allow us to manage this and reproduce the expected behaviour.

aawdls (approved):
Thanks for the modifications.

PR #41, opened 2023-07-18
Add widgets

abigailalexander:
This PR adds extra widgets
- Arc
- BoolButton
- ByteMonitor
- Ellipse
- Polygon

and also includes tests for each widget. For the BoolButton widget, it is intended to have PV write access when clicked. For now this has not been implemented, but will need to be included in future.
I tested these widgets by implementing the individual widget code in an older version of cs-web-proto that did not use the cs-web-lib library, which is not ideal but I couldn't get the `npm link` method to work. I could not see any issues with the widget functionalities when testing them.


abigailalexander:
Thanks for the feedback Becky! Hopefully the new commits should fix these issues:

- The inverted arc should not appear the correct way around (this did take a while to do, so let me know if there are still issues as I am not 100% sure it is completely fixed)
- There is now an LED in the BoolButton - I had initially not included it because I was unsure of its purpose and whether it was used at Diamond but I feel that even if it is not used it is useful to have. The LED should hopefully change colour when the button is clicked.
- The border around the ellipse has been removed when line width is 0

I have also updated the appearance of ByteMonitor and adjusted the sizing of a few other widgets to be more accurate to CSStudio, which hopefully all behave as expected.

rjwills28:
Thanks for the changes Abii. Just a few more comments:
1. The Arc widget is no longer inverted and looks correct for all angle except when I just tried putting in a negative angle (e.g. -90). CS-Studio supports this but I think it crashes in the browser and the widget disappears. 
2. I can see the LED light up when I click on the Boolean Button and I know it doesn't currently have write access to update the connected PV but I did notice that the LED does not change when I `caput` to the PV to change it, i.e. it doesn't seem to reflect the actual state of the PV?
3. I also notice on the Boolean Button that the label comes up when you click on the button (e.g. ON, OFF). In CS-Studio the showing of the label is configurable with a property (`show_boolean_label`) so I wonder whether we should do the same here?

abigailalexander:
Hopefully the issues you raised should now be fixed with these latest commits. I have also added write PV ability to the BoolButton, although I didn't seem to be able to test this fully with the local coniql instance I was running due to errors.

rjwills28 (approved):
Thanks for all the fixes Abii. I've retest and it's all working as expected. I was able to test the write ability of the BoolButton using my local Coniql and that now works nicely (to do this you have to start Coniql with the `--cors` argument to allow 'cross origin resource sharing' which will then allow you to write, e.g. `coniql --cors`). So all looks good.

PR #42, opened 2023-08-16
Update bytemonitor widget

abigailalexander:
Updated the widget so that the sizing of LEDs now more accurately replicates CSStudio sizing. I have also improved the 3D effect appearance and updated tests.
This is their appearance now in cs-web-lib
Bytemonitor in cs-web-lib
Bytemonitor in cs-web-lib
versus how the same widgets look in CSStudio
Bytemonitor in CSStudio
Bytemonitor in CSStudio

### DIFF
--- a/src/ui/widgets/ByteMonitor/byteMonitor.test.tsx
+++ b/src/ui/widgets/ByteMonitor/byteMonitor.test.tsx
@@ -26,11 +26,11 @@ describe("<ByteMonitorComponent />", (): void => {
     const bits = byteMonitor.children as Array<ReactTestRendererJSON>;
     expect(bits.length).toEqual(16);
 
-    expect(bits[0].props.style.marginRight).toEqual("-3px");
-    expect(bits[1].props.style.borderWidth).toEqual(0);
+    expect(bits[0].props.style.marginRight).toEqual("-2px");
+    expect(bits[1].props.style.borderWidth).toEqual("2px");
     expect(bits[5].props.style.backgroundColor).toEqual("rgba(0,100,0,255)");
-    expect(bits[10].props.style.boxShadow).toEqual(
-      "inset 0.0625px 0.0625px 0.1px rgba(255,255,255,.5), 1px 1px white, -1px -1px darkgray"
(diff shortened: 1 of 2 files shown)


AlexanderWells-diamond:
That now looks much better. Saves us having to tweak a lot of screens to make alignments work.

I note that there is a difference between the WebUI and CSS version: The text boxes contain 0's whereas the CSS textboxes are blank. Is this a deliberate change in the behaviour of textboxes?

abigailalexander:
I'm not sure if it's a deliberate change in behaviour as I didn't write the original code, but it looks like if the `format_type` field for TextUpdate is set to "default" and the value is "" this gets coerced into a double, which causes the 0 to appear by this section of code in readback.tsx: 
```
if (value.display.choices) {
      // Enum PV so use string representation.
      displayedValue = DType.coerceString(value);
    } else if (prec !== undefined && !isNaN(DType.coerceDouble(value))) { // this evaluates to true
      if (formatType === "exponential") {
        displayedValue = DType.coerceDouble(value).toExponential(prec);
      } else {
        displayedValue = DType.coerceDouble(value).toFixed(prec); // here "" is converted to 0
      }
    } else if (formatType === "string") {
      const valarr = value.getArrayValue();
      if (valarr !== undefined) {
        displayedValue = DType.byteArrToString(valarr);
      } else {
        displayedValue = DType.coerceString(value);
      }
    } else {
      displayedValue = DType.coerceString(value);
    }
```
If this behaviour is not what is wanted and not deliberate I could look into changing it in future.

rjwills28:
Yes it looks like the `coerceDouble` function is not quite working properly. On line 203, see https://github.com/dls-controls/cs-web-lib/blob/d7a71b3c72282d42eb0f0872a7a08ceda5fd3c1a/src/types/dtypes.ts#L203 it checks if the string value is undefined but should probably also check that it is not empty before trying to parse it as a `Number` because `Number("")` will return `0`. I think this was probably an oversight and so should be fixed.

Otherwise I've run a few test locally and all looks good to me.

rjwills28 (approved):

Issue #47, opened 2024-01-24
Line widget appearance is incorrect


abigailalexander:
The Line widget does not behave as seen in Phoebus and CSS. In Phoebus and CSS, it is composed of multiple points (x, y coordinates) that determine each section of the line. In cs-web-lib this is ignored and the line is treated as a straight line across the width of the widget. 
Additionally, there is the option to add arrows to the Line in CSS/Phoebus. This is missing in cs-web-lib.

Screenshots show the Line widget in cs-web-lib and Phoebus.
![Screenshot from 2024-01-24 10-48-35](https://github.com/dls-controls/cs-web-lib/assets/91315856/828411d6-1739-4e2a-a80f-3275324ef7fd)
![Screenshot from 2024-01-24 10-49-13](https://github.com/dls-controls/cs-web-lib/assets/91315856/b3ed423e-e7b3-481b-a089-58dbd82d5588)

abigailalexander:
Fixed by PR #49

PR #49, opened 2024-01-25
Fix Line widget Issue #47

abigailalexander:
Fixed the functionality of the line widget so the behaviour more closely matches that of CSS/Phoebus

- Line is an svg `polyline` with multiple points of x,y coordinates joined by lines
- Arrowheads can be added to either or both ends of the line
- Tests updated, along with snapshot

### DIFF
--- a/src/ui/widgets/Line/line.tsx
+++ b/src/ui/widgets/Line/line.tsx
@@ -6,19 +6,22 @@ import {
   FloatPropOpt,
   ColorPropOpt,
   BoolPropOpt,
-  FloatProp
+  FloatProp,
+  PointsPropOpt
 } from "../propTypes";
 import { registerWidget } from "../register";
-import { ShapeComponent } from "../Shape/shape";
 import { Color } from "../../../types/color";
+import { Point } from "../../../types/points";
 
 const LineProps = {
   width: FloatProp,
+  height: FloatProp,
   lineWidth: FloatPropOpt,
   backgroundColor: ColorPropOpt,
   visible: BoolPropOpt,
   transparent: BoolPropOpt,
-  rotationAngle: FloatPropOpt
+  rotationAngle: FloatPropOpt,
+  points: PointsPropOpt
 };
 
 export type LineComponentProps = InferWidgetProps<typeof LineProps> &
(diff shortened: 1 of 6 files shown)


rjwills28:
I have tested the `Polyline` widget in CSS and the web UI to compare. The line now has multiple joined points and I can display an arrow head correctly .
I noticed 2 slight issues with the arrow head in the web UI which differs in behaviour from CSS:
1. In CSS there is the option to `Fill Arrow` (yes, no). 'Yes' works the same in both but the web UI does not render the no fill option, see below with no fill for both:
css_arrow_nofill web_arrow_nofill
2. If the `Line Width` property is greater than 1 then the arrow seems to get much larger than it should in the web UI, e.g. with a `Line Width` of 5:
css_arrow_width5 web_arrow_width5

src/ui/widgets/Line/line.tsx
```diff
@@ -28,30 +33,95 @@ export const LineComponent = (props: LineComponentProps): JSX.Element => {
...
-  };
+  const transform = `rotation(${rotationAngle},0,0)`;
 
-  const shapeProps = {
-    shapeWidth: `${width}px`,
-    shapeHeight: `${lineWidth}px`
-  };
+  // Each marker definition needs a unique ID or colours overlap
+  // Get random decimal number, take decimal part and truncate
+  const uid = String(Math.random()).split(".")[1].substring(0, 6);
```
rjwills28, line 50: There is an npm library that could do this called uuid: https://www.npmjs.com/package/uuid. Looks fairly easy to use. Not that there is anything wrong with this method, just wondering if using the library would make it more transparent although it would add another dependency.

abigailalexander: I was hesitant to add another dependency, but I think it's definitely worth it for transparency. I've added that in now.

abigailalexander:
The latest commits should add in the fill arrow functionality. 

1. The arrow should now appear unfilled like seen in CSS/Phoebus when `fillArrow` is false. The behaviour of the filled/unfilled arrows is slightly different so I've attempted to capture that behaviour.
2. The linewidth of the unfilled arrow scale directly with `lineWidth`, whereas for the filled arrow the arrow size remains the same when line width is adjusted.

Additionally, the account for the case where the `lineWidth` is wider than the filled arrow and this obscures the point of the arrow, when `fillArrow` is true the length of the line segment is reduced. This is necessary because of how SVG places the markers, although is a little confusing to explain.

src/ui/widgets/Line/line.tsx
```diff
@@ -40,57 +42,100 @@ export const LineComponent = (props: LineComponentProps): JSX.Element => {
...
     arrowLength = 2,
-    arrows = 0
+    arrows = 0,
+    fillArrow
   } = props;
 
   const transform = `rotation(${rotationAngle},0,0)`;
 
   // Each marker definition needs a unique ID or colours overlap
   // Get random decimal number, take decimal part and truncate
```
rjwills28, line 52: Remove this line of the comment.

rjwills28 (approved):
Thanks for making the changes. I've tested and all looks good now - behaviour resembles CS-Studio. One change mentioned above to remove out-dated comment, but otherwise approved.

abigailalexander:
Thanks for looking over it, I've removed that out-dated comment now.

PR #50, opened 2024-01-29
Change EmbeddedDisplay overflow mode to "auto"

abigailalexander:
Changed the overflow behaviour for EmbeddedDisplay to "auto". This means that scrollbars appear when necessary, but do not appear when there is no overflow. 

This was after reported issues in Google Chrome of seeing double scrollbars even when application was in full-screen mode. Firefox hides scrollbars when not needed, so there is no difference in behaviour for this browser.


rjwills28 (approved):
Tested on the machine-status pages in Chrome and can confirm this fixes the issue.

PR #54, opened 2024-04-09
Add choice widget

abigailalexander:
Added a Choice Button widget. I've tested the itemsFromPv and enabled/disabled functionality, as well as writing to PV and these all seem to work. I modelled this off the Phoebus version of the widget


rjwills28 (commented):
I've tested this widget on a BOB and OPI file and all looks good!
Code looks fine as well. There are just 2 conflicts to resolve on the parsers after the bobParser PR was merged into main. I think the resolution is to keep both changes which are just introducing new methods in the same location (but do double check me).

abigailalexander:
Great, thanks Becky! I'll resolve those conflicts when I merge it in

PR #56, opened 2024-06-04
Fix bob widget issues

abigailalexander:
This PR fixes various issues with widgets that I noticed when using .bob files:

- Added `image_stretch` field (this is `stretch_to_fit` in .opi files)
- Modified bobParseFonts to use strings instead of numbers as keys
- Added a check for whether a .opi or .bob file is used for rules, as .opi uses `pv_name` and .bob uses `pv`
- Changed default background and text colour for Label and Readback to match with Phoebus defaults


src/ui/widgets/EmbeddedDisplay/bobParser.ts
```diff
@@ -245,10 +245,10 @@ export function parseBob(
       (pvName: ElementCompact): PV => opiParsePvName(pvName, defaultProtocol)
     ],
     font: ["font", bobParseFont],
-    rules: [
-      "rules",
-      (rules: Rule[]): Rule[] => opiParseRules(rules, defaultProtocol)
-    ],
+    // rules: [
```
rjwills28, line 248: Can we remove this commented out code now?

src/ui/widgets/Readback/readback.module.css
```diff
@@ -5,11 +5,11 @@
   white-space: nowrap;
   text-overflow: ellipsis;
   background-color: #383838;
```
rjwills28, line 7: The background colour for the readback widget is currently very dark, whereas the default Phoebus background is a lighter grey - I believe the `background-color` should be `#f0f0f0`.
Current: 
current_readback
Proposed:
readback_new

rjwills28 (commented):
Thanks for these changes. I've just added a few comments/suggestions.

src/ui/widgets/EmbeddedDisplay/bobParser.ts
abigailalexander: Ah yes, I meant to remove that. Thank you for spotting it

src/ui/widgets/Readback/readback.module.css
abigailalexander: Thanks for the spot, changed this now.

rjwills28 (approved):

PR #59, opened 2024-07-15
Replaced dls-controls with DiamondLightSource

garryod:
Repository was automatically transferred from `dls-controls/cs-web-lib` using https://gitlab.diamond.ac.uk/github/github-scripts

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@ An npm library for Control Systems web applications
 
 ## Installation
 Install via npm:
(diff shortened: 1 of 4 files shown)


coretl:
@abigailalexander we've moved the repo, but not sure which of these changes need to be made to the repo as a result, please can you edit and merge this?

abigailalexander (approved):
NPM organisation names need to be all lowercase, so I've corrected that. Every other change is correct, so I'll merge

PR #60, opened 2024-08-22
Update symbol widget

abigailalexander:
Current implementation of the Symbol widget in cs-web-lib was based off the MultiStateMonitor widget in CSStudio. The Phoebus Symbol widget implementation was not applied.

I've now updated this widget to behave like the Phoebus widget. Tests and snapshots have also been updated. We check whether the symbol should behave like MultiStateMonitor or Symbol, so both opi and bob file behaviours are preserved.

### DIFF
--- a/src/ui/widgets/EmbeddedDisplay/bobParser.ts
+++ b/src/ui/widgets/EmbeddedDisplay/bobParser.ts
@@ -51,7 +51,8 @@ const BOB_WIDGET_MAPPING: { [key: string]: any } = {
   progressbar: "progressbar",
   rectangle: "shape",
   choice: "choicebutton",
-  scaledslider: "slidecontrol"
+  scaledslider: "slidecontrol",
+  symbol: "symbol"
 };
 
 // Default width and height of widgets in Phoebus
@@ -75,7 +76,8 @@ export const WIDGET_DEFAULT_SIZES: { [key: string]: [number, number] } = {
   polyline: [100, 20],
   progressbar: [100, 20],
   rectangle: [100, 20],
-  scaledslider: [400, 55]
+  scaledslider: [400, 55],
+  symbol: [100, 100]
 };
 
 function bobParseType(props: any): string {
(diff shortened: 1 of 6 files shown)


rjwills28:
I have tested this with both an OPI and BOB file using a `MultiStateSymbolMonitor` and a `Symbol` widget respectively. The CSS `MultiStateSymbolMonitor` is still displayed correctly and now the Phoebus `Symbol` is correctly displayed (instead of getting an error of unknown widget). 

Two things I did notice were related to the rotation property:
1. In the OPI file, the rotation is applied (`<degree>45</degree>`) but it is cut off as it doesn't fit in the box it is rendered in - see image below with cs-web-lib left and cs-studio right.
Screenshot from 2024-12-06 15-30-34
2. No rotation is applied to the image from the BOB file when specified, i.e. with `<rotation>45.0</rotation>`. I guess this property is just ignored as it has a different tag name to the CSS property.

Not sure whether we want to fix this here or as a separate issue? Perhaps depends on how complicated it is as this PR certainly addresses the implementation of the BOB symbol.

abigailalexander:
I hadn't noticed those issues, thank for pointing them out! I'll have a look into them, and if they seem to be a separate issue that would be difficult to fix, I can do it as a separate PR.

src/ui/widgets/EmbeddedDisplay/bobParser.ts
```diff
@@ -210,8 +210,8 @@ function bobParseResizing(jsonProp: ElementCompact): string {
 
 function bobParseSymbols(jsonProp: ElementCompact): string[] {
   const symbols: string[] = [];
-  jsonProp["symbol"].forEach((item: any) => {
-    symbols.push(item._text);
+  Object.values(jsonProp["symbol"]).forEach((item: any) => {
+    symbols.push(item);
```
rjwills28, line 214: I am getting a parsing error with this change, which I wasn't getting before.
```
Error: Failed widget description type: Invalid widget description `symbols[0]` of type `object` supplied to `Symbol$1`, expected `string`.
```
It worked when I put it back to `item._text`?

rjwills28 (changes requested):

src/ui/widgets/EmbeddedDisplay/bobParser.ts
abigailalexander: Thanks for spotting this! I've added an extra check to see if we receive an object or a string and parse it accordingly, to account for single symbol and multiple symbol widgets.

rjwills28 (approved):
Look great - thanks for identifying that that was the issue.

PR #61, opened 2024-09-04
Update to use Vite instead of create-react-app

abigailalexander:
Create-react-app is now deprecated, and came with many vulnerabilities in its dependencies. The recommended tool to use now is vite. I have updated cs-web-lib to use this, as well as:

- Swapped from jest to vitest
- Updated the github actions to use Node 18. This is necessary because create-react-app used polyfills to provide methods like fetch globally, and vite does NOT do the same. However, Node 18 onwards provides fetch globally. We are also currently on Node 21, so 16 is outdated. 
- Changed the name of environment variables to match vite conventions. This changes from process.env.REACT_APP_(NAME) to import.metadata.env.VITE_(NAME).
- Minor linting changes


.eslintrc.json
iamvigneshwars, line 1: EOF

.github/workflows/node.install.yml
```diff
@@ -14,13 +14,13 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [16.x]
+        node-version: [18.x]
```
iamvigneshwars, line 17: Any specific reason not using the latest version (22.x) ?

.github/workflows/node.install.yml
```diff
@@ -14,13 +14,13 @@ jobs:
...
         # See supported Node.js release schedule at https://nodejs.org/en/about/releases/
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v4
```
iamvigneshwars, line 21: It's better to use a fully qualified version `actions/checkout@v4.1.7`

tsconfig.json
iamvigneshwars, line 1: EOF

iamvigneshwars (commented):
There are two EOF and few suggestions in the CI, apart from that LGTM!

.github/workflows/node.install.yml
abigailalexander: Mostly just because I was still running Node 18 on my desktop, but I updated to 22 and everything is fine so I've updated the CI to use 22 as well now.

iamvigneshwars (commented):
Since you are updating stuffs, it is also worth setting up a dependabot in the CI. It keeps all the libraries up-to date. It will open a PR for every version upgrade so you dont have to worry about any new breaking changes.

abigailalexander:
Good idea, although I'll probably do that in a separate PR.

iamvigneshwars (approved):

PR #63, opened 2024-09-19
Fix vite issues

abigailalexander:
After migrating to use vite the other day, I noticed a few issues when utilising cs-web-lib in machine-status:

- Previously added `"type": "module"` into package.json. This has now been removed because it isn't needed and causes an error when running tests in machine-status:
```
ReferenceError: exports is not defined in ES module scope
This file is being treated as an ES module because it has a '.js' file extension and '/node_modules/@diamondlightsource/cs-web-lib/package.json' contains "type": "module". To treat it as a CommonJS script, rename it to use the '.cjs' file extension.
```
-  Vite exposes environment variables on `import.meta.env` instead of `process.env` so I had previously changed this over, but in order to maintain compatibility and allow client applications still using cra to use cs-web-lib, I've modified it so that both are used. We attempt to use `process.env` first and if that is empty, fall back to `import.meta.env`

The changes in the PR should correct these issues.

### DIFF
--- a/src/redux/store.ts
+++ b/src/redux/store.ts
@@ -8,10 +8,14 @@ import { SimulatorPlugin } from "../connection/sim";
 import { ConiqlPlugin } from "../connection/coniql";
 import { ConnectionForwarder } from "../connection/forwarder";
 
-const CONIQL_SOCKET = import.meta.env.VITE_CONIQL_SOCKET;
-const CONIQL_SSL = import.meta.env.VITE_CONIQL_SSL === "true";
+const CONIQL_SOCKET =
+  process.env.VITE_CONIQL_SOCKET ?? import.meta.env.VITE_CONIQL_SOCKET;
+const CONIQL_SSL =
+  (process.env.VITE_CONIQL_SSL ?? import.meta.env.VITE_CONIQL_SSL) === "true";
 const THROTTLE_PERIOD = parseFloat(
-  import.meta.env.VITE_THROTTLE_PERIOD ?? "100"
+  process.env.VITE_THROTTLE_PERIOD ??
+    import.meta.env.VITE_THROTTLE_PERIOD ??
+    "100"
 );
 
 const simulator = new SimulatorPlugin();
--- a/package.json
+++ b/package.json
@@ -3,7 +3,6 @@
   "version": "0.3.0",
   "description": "Control system web library",
   "main": "dist/cjs/index.js",
-  "type": "module",
   "scripts": {
     "rollup": "rollup -c",
     "test": "vitest",
--- a/src/misc/profilerCallback.ts
+++ b/src/misc/profilerCallback.ts
@@ -8,7 +8,9 @@ import log from "loglevel";
  * and for more reliable measurements
  */
 
-const PROFILE_ENABLED = import.meta.env.VITE_PROFILER_ENABLED === "true";
+const PROFILE_ENABLED =
+  (process.env.VITE_PROFILER_ENABLED ??
+    import.meta.env.VITE_PROFILER_ENABLED) === "true";
 const ITEMS_TO_STORE = 100;
 
 class Timings {
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -22,7 +22,10 @@
     "noFallthroughCasesInSwitch": true,
     "resolveJsonModule": true,
     "isolatedModules": true,
-    "noEmit": true
+    "noEmit": true,
+    "types": [
+      "vite/client"
+    ],
   },
   "include": [
     "src"


src/redux/store.ts
```diff
@@ -8,10 +8,14 @@ import { SimulatorPlugin } from "../connection/sim";
 import { ConiqlPlugin } from "../connection/coniql";
 import { ConnectionForwarder } from "../connection/forwarder";
 
-const CONIQL_SOCKET = import.meta.env.VITE_CONIQL_SOCKET;
-const CONIQL_SSL = import.meta.env.VITE_CONIQL_SSL === "true";
+const CONIQL_SOCKET =
+  process.env.VITE_CONIQL_SOCKET ?? import.meta.env.VITE_CONIQL_SOCKET;
```
rjwills28, line 12: I am having problems getting this to work with a client still running cra. I get the following error after issuing `npm run start`:
```
Failed to compile.

./node_modules/@diamondlightsource/cs-web-lib/dist/esm/index.js 19497:105
Module parse failed: Unexpected token (19497:105)
File was processed with these loaders:
 * ./node_modules/babel-loader/lib/index.js
You may need an additional loader to handle the result of these loaders.
| }
| var _a$1, _b, _c, _d;
> const CONIQL_SOCKET = (_a$1 = process.env.VITE_CONIQL_SOCKET) !== null && _a$1 !== void 0 ? _a$1 : import.meta.env.VITE_CONIQL_SOCKET;
| const CONIQL_SSL = ((_b = process.env.VITE_CONIQL_SSL) !== null && _b !== void 0 ? _b : import.meta.env.VITE_CONIQL_SSL) === "true";
| const THROTTLE_PERIOD = parseFloat((_d = (_c = process.env.VITE_THROTTLE_PERIOD) !== null && _c !== void 0 ? _c : import.meta.env.VITE_THROTTLE_PERIOD) !== null && _d !== void 0 ? _d : "100");
```
I think this comes down to it not being able to parse `import.meta.env.*` ? Is this something you were able to workaround?

rjwills28 (changes requested):
I couldn't get the 'original' version of machine-status (using cra) working with the library because of a parse fail (see comment). Even if I removed the references to `import.meta.env` and just leave `process.env.VITE_CONIQL_SOCKET`, the machine-status app cannot read the `VITE_CONIQL_SOCKET` environment variable as cra requires the variables to have the prefix `REACT_APP_` so I'm not sure this is going to work (unless I have missed something?). 

Another alternative I've found would be to leave all of the environment variables in cs-web-lib in the form of `process.env.REACT_APP_` as they were (i.e. not using `import.meta.env`) e.g.
```
const CONIQL_SOCKET = process.env.REACT_APP_CONIQL_SOCKET;
```

This means it will work as normal with the original. 

To work with clients using 'vite' we can add a definition to the `vite.config.ts` to get it to use environment variables in the form `process.env` (see https://dev.to/whchi/how-to-use-processenv-in-vite-ho9).

In my client that I have ported to vite I add the following to the `vite.config.ts` file: 
```
define: {
    'process.env.REACT_APP_CONIQL_SOCKET': JSON.stringify(env.REACT_APP_CONIQL_SOCKET),
    'process.env.REACT_APP_CONIQL_SSL': JSON.stringify(env.REACT_APP_CONIQL_SSL),
    'process.env.REACT_APP_PROFILER_ENABLED': JSON.stringify(env.REACT_APP_PROFILER_ENABLED),
    'process.env.REACT_APP_THROTTLE_PERIOD': JSON.stringify(env.REACT_APP_THROTTLE_PERIOD),
    'process.env.REACT_APP_LOG_LEVEL': JSON.stringify(env.REACT_APP_LOG_LEVEL),
    'process.env.REACT_APP_CYCLE_CONFIG_FILE': JSON.stringify(env.REACT_APP_CYCLE_CONFIG_FILE),
    'process.env.REACT_APP_BUILD_TARGET': JSON.stringify(env.REACT_APPs_BUILD_TARGET),
},
```
Then the .env files can be left as is in the form:
```
REACT_APP_CONIQL_SOCKET=localhost:8080
REACT_APP_CONIQL_SSL=false
```

This is the only real way I can find to make the library compatible with both cra and vite, unless I have missed something in my test setup?? My only worry with this solution is that it sort of compromises the port to vite by retaining react-like env variables.

src/redux/store.ts
abigailalexander: I thought it had worked when I tested previously, but I just tried again now and encountered this same issue. I don't think there is a workaround for this.

abigailalexander:
I must have been mistaken that it worked when I tested it before, perhaps I still had some vite config floating around that enabled it. I think there are two solutions we can go with:

- Keep as is, and put a note saying this doesn't work with CRA. Given it is now considered deprecated, this isn't too big of an issue as no projects should start using it nowadays. However, I don't know if cs-web-lib will work for any bootrstapping other than vite (if any others exist)....?
- Use the `define` configuration in vite clients to convert back to `process.env` variables. This creates more work for vite clients but allows CRA compatibility still. Most of the benefits from using vite are reduced dependency vulnerabilities and a still-maintained package, so choosing this option doesn't negate the benefits of using vite.

The second one seems like it would be a slightly tidier option and it wouldn't be too much effort to add a few lines to the docs about setup. Thoughts?

rjwills28:
I was building different versions of machine-status against different versions of cs-web-lib and did get a bit confused as to what worked and what didn't (went round it a few times).

I am in two minds about which option is best. If we allow for cs-web-lib to run with both vite and cra applications we can essentially 'put off' updating the clients but still get the security benefits from the library running vite. However, do the clients themselves not _need_ to also be updated to overcome security vulnerabilities. I guess I'm wondering if we plan to just update all the clients anyway then there is no need to maintain backwards compatibility. Who else is currently using the library - machine-status, cs-web-proto ??

rjwills28:
Our current version of machine-status only builds and runs on node 16 (which is no longer supported) so if we were to update the node version (for support and security reasons) we would have to make updates to the client as well anyway.

abigailalexander:
I spoke to @MJGaughran  and he thinks that not supporting CRA anymore is the right decision. The only uses of cs-web-lib currently are machine status, cs-web-proto and someone from CEA has recently started experimenting with it, though I'm not sure if he used vite to create the project. Machine status and cs-web-proto will also need updating to vite yes, I started this process with machine status and it isn't too time consuming.

I'll add into my docs PR that create-react-app will no longer be supported then. I believe then that this PR can remain as-is? (as the `process.env` is required for vitest - although there is an alternative solution we could implement in machine status to remove the need for this, but I couldn't get it working. Mentioned in this issue https://github.com/vitest-dev/vitest/issues/2970)

rjwills28 (approved):
Yes, agree with that decision and this PR is ready to be merged as is.

PR #65, opened 2024-10-22
Fix state update on an unmounted component

rjwills28:
After updating the machine-status client to use `vite` and `vitest` I noticed that I was getting some warnings when running the tests:
```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
    at EmbeddedDisplay (/home/user/cs-web/machine-status/node_modules/@diamondlightsource/cs-web-lib/dist/cjs/index.js:54548:42)
    at LoadEmbeddedDirect (/home/user/cs-web/machine-status/src/app.tsx:39:29)
    at RedirectAfterTimeout (/home/user/cs-web/machine-status/src/app.tsx:618:1)
    at LoadRedirectView
    at Route (/home/user/cs-web/machine-status/node_modules/react-router/cjs/react-router.js:664:29)
    at Switch (/home/user/cs-web/machine-status/node_modules/react-router/cjs/react-router.js:866:29)
    at div
    at Provider (/home/user/cs-web/machine-status/node_modules/react-redux/lib/components/Provider.js:21:20)
    at AppScreen (/home/user/cs-web/machine-status/src/app.tsx:105:22)
    at Router (/home/user/cs-web/machine-status/node_modules/react-router/cjs/react-router.js:283:30)
    at BrowserRouter (/home/user/cs-web/machine-status/node_modules/react-router-dom/cjs/react-router-dom.js:75:35)
```
As it states, it is due to it trying to update the state on an unmounted component. This might not be an issue in production but is highlighted in the tests as the component gets unmounted almost immediately after mounting.

To eliminate this I have simply added a check that the component is mounted before updating the state and then have added a return function which is called when the component is unmounted and can set the `isMounted` check variable to `false`.

### DIFF
--- a/src/ui/widgets/EmbeddedDisplay/useOpiFile.ts
+++ b/src/ui/widgets/EmbeddedDisplay/useOpiFile.ts
@@ -78,14 +78,14 @@ export function useOpiFile(file: File): WidgetDescription {
   const [contents, setContents] = useState<WidgetDescription>(EMPTY_WIDGET);
 
   useEffect(() => {
-    let isMounted = false;
+    let isMounted = true;
     const fetchData = async (): Promise<void> => {
       if (fetchPromises.hasOwnProperty(file.path)) {
         // This resource has been requested; once the cached
         // promise has been resolved the fileCache should be
         // populated.
         await fetchPromises[file.path];
-        if (!isMounted) {
+        if (isMounted) {
           setContents(fileCache[file.path]);
         }
       } else {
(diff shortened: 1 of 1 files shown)


abigailalexander (approved):
Tested and can confirm this removes the error when running vitest tests in machine-status, and affects no other functionality.

PR #66, opened 2024-11-01
Switch library to use pvws backend

rjwills28:
I have replaced Coniql with PVWS which includes:
- updating the code used to connect to pvws through a websocket and how to parse the data from pvws. 
- adding tests for the pvws connection using vitest.
- updating the .env parameters to refer to `VITE` instead of `CONIQL`. This will subsequently need to be updated in all clients.


src/connection/pvws.test.ts
```diff
@@ -0,0 +1,97 @@
...
+
+describe("PvwsPlugin", (): void => {
+  let cp: PvwsPlugin;
+  let mockConnUpdate: jest.Mock;
+  let mockValUpdate: jest.Mock;
+  let ws: WS;
+  beforeEach(async () => {
+    ws = new WS("ws://a.b.c:100/pvws/pv");
+    cp = new PvwsPlugin("a.b.c:100", false);
+    mockConnUpdate = jest.fn();
```
abigailalexander, line 13: Is there a reason we use jest.fn() to mock here, instead of vi.fn()?

rjwills28: Ah no, I think I just missed that. Thanks for catching that - I'll fix and retest.

abigailalexander (approved):
I tested in machine-status, works as expected.

PR #69, opened 2024-11-28
Export some hooks and necessary datatypes

### LINKED ISSUE #68: Expose the hooks
As a developer of GUIs for experimental use I'd like to make use of the communication utilities in cs-web-lib for my own widgets and components, so I would like hooks like `useDevice`, `useConnection` etc. to be exported

dperl-dls:
Fixes #68 

I think this is enough to get us going, thanks for being open to these changes!

We might want to add a few more later but I'm erring on the side of caution for the moment. We are actually kind of interested in splitting out the PVWS communication + React hooks into a separate library (which we would be willing to help maintain!) because it could be reused in a lot of projects which don't correspond to CS-Studio stuff, but that is a discussion to have with @DominicOram next year.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "@diamondlightsource/cs-web-lib",
-      "version": "0.4.0",
+      "version": "0.5.0",
       "license": "ISC",
       "dependencies": {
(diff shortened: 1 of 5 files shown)


abigailalexander (approved):
Looks all good to me.

PR #70, opened 2024-12-10
Preserve 'use-client' directives during rollup

rjwills28:
Previously we were getting a lot of warnings thrown during `rollup` that the `use client` directive from the 'mui' package were ignored:
```
(!) node_modules/@mui/styled-engine/GlobalStyles/GlobalStyles.js (1:0): Module level directives cause errors when bundled, "use client" in "node_modules/@mui/styled-engine/GlobalStyles/GlobalStyles.js" was ignored.
/cs-web-lib/node_modules/@mui/styled-engine/GlobalStyles/GlobalStyles.js:1:0
1: 'use client';
   ^
2: 
3: import * as React from 'react';
```

This `use client` directive is used by the mui package to mark the module as client code and therefore should probably be preserved during the rollup. There is a package `rollup-preserve-directives` that can do this so I have added this and configured it in the rollup.config file.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -53,6 +53,7 @@
         "rollup-plugin-dts": "^6.1.1",
         "rollup-plugin-peer-deps-external": "^2.2.4",
         "rollup-plugin-postcss": "^4.0.2",
+        "rollup-preserve-directives": "^1.1.3",
         "typescript": "^5.6.3",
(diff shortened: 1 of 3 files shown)


abigailalexander (approved):
Tested that the rollup warnings are gone and the package installs as usual, all works. LGTM!

PR #71, opened 2024-12-10
Fix findDOMNode deprecation warning in slideshow widget

rjwills28:
This warning is currently thrown when running the tests:
```
Warning: findDOMNode is deprecated and will be removed in the next major release. Instead, add a ref directly to the element you want to reference. Learn more about using refs safely here: https://reactjs.org/link/strict-mode-find-node
    at Transition (/cs-web-lib/node_modules/react-transition-group/cjs/Transition.js:135:30)
    at CSSTransition (/cs-web-lib/node_modules/react-transition-group/cjs/CSSTransition.js:120:35)
    at SwitchTransition (/cs-web-lib/node_modules/react-transition-group/cjs/SwitchTransition.js:159:35)
    at SwitchableWidget (/cs-web-lib/src/ui/widgets/Slideshow/slideshow.tsx:49:25)
    at div
    at div
    at SlideshowComponent (/cs-web-lib/src/ui/widgets/Slideshow/slideshow.tsx:86:61)
```

I've followed the advice to pass a `ref` directly into the element and `CSSTransition` component. Also see discussion here with a similar suggestion/implementation: https://github.com/reactjs/react-transition-group/issues/668.

Beyond this, I'm actually not sure where the slideshow widget is used? Maybe it can actually be removed...

### DIFF
--- a/src/ui/widgets/Slideshow/slideshow.tsx
+++ b/src/ui/widgets/Slideshow/slideshow.tsx
@@ -31,10 +31,15 @@ export const SwitchableWidget = (props: {
   index: number;
   children: [ReactElement];
   transition: { readonly [key: string]: string };
+  nodeRef: React.Ref<HTMLDivElement>;
 }): JSX.Element => {
+  // react-transition-group internally uses findDOMNode which
+  // is deprecated. Fix by passing nodeRef which points to the
+  // transitioning child.
   return (
     <SwitchTransition mode="out-in">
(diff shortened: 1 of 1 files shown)


abigailalexander (approved):
Can confirm this fixes the warning. 
I'm also not sure what the use of the Slideshow widget is, as it isn't part of css/phoebus. I agree we should look into whether it could be removed in future

PR #74, opened 2025-01-30
Update font sizing

abigailalexander:
Previously, font sizing in cs-web-lib used the [62.5% font-size trick](https://www.aleksandrhovhannisyan.com/blog/62-5-percent-font-size-trick/) for font-sizing, which requires client applications to set the root font size to 62.5% for fonts to appear the correct size. This isn't very transparent and there seem to be mixed reviews on using it. It causes issues when trying to use cs-web-lib in a client application that uses e.g. MUI components alongside bob screens.

I have swapped the font back to calculating REM from pixels by using the browser default font size of 16px and added a comment for context. This won't scale correctly if a browser user manually adjusts the browser default font size, but the .bob screens don't scale correctly anyway.

### DIFF
--- a/src/types/font.test.ts
+++ b/src/types/font.test.ts
@@ -6,7 +6,7 @@ describe("Font", (): void => {
     const fontStyle = font.css();
     expect(fontStyle).toEqual({
       fontFamily: "sans,sans-serif",
-      fontSize: "1rem",
+      fontSize: "0.625rem",
       fontWeight: "normal",
       fontStyle: "normal"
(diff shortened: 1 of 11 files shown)


rjwills28 (approved):
Thanks for these changes. I tested against cs-web-proto which has a test widget page and noticed a few other sizing issues so I've just fixed the default font sizes for a few additional widgets.
Hopefully we've caught everything - otherwise we'll just have to fix them when we come across them.

PR #75, opened 2025-02-04
Add bobParseActions

abigailalexander:
Actions in bob files were previously parsed by `opiParseActions`, which didn't work due to small differences between actions in opi and bob files. These include:

- The `type` of action is lowercase in bob i.e. "open_display" but uppercase in opi
- bob files use the `target` property to determine if the new screen opens in a tab or in place of the current screen, rather than mode and position
- The path to the file to open is given by the `file` property rather than the `path` property in bob files

I created `bobParseActions` to correctly parse with these differences

### DIFF
--- a/src/ui/widgets/EmbeddedDisplay/bobParser.ts
+++ b/src/ui/widgets/EmbeddedDisplay/bobParser.ts
@@ -250,9 +250,12 @@ export function bobParseActions(
       case "replace":
         return "main";
       case "tab":
-        return "details";
+        // If a named tab is given, open there
+        // Otherwise default to main
+        if (action.name) return action.name._text;
+        return "main";
       default:
-        return "details";
+        return "main";
     }
   };
 
--- a/src/ui/widgets/EmbeddedDisplay/bobParser.ts
+++ b/src/ui/widgets/EmbeddedDisplay/bobParser.ts
@@ -23,6 +23,7 @@ import {
 import { PV } from "../../../types/pv";
 import { OpiFile, Rule } from "../../../types/props";
 import {
+  OPEN_PAGE,
   OPEN_TAB,
   OPEN_WEBPAGE,
   WidgetActions,
@@ -282,8 +283,9 @@ export function bobParseActions(
           }
         });
       } else if (type === "open_display") {
+        const type = action.target._text === "replace" ? OPEN_PAGE : OPEN_TAB;
         processedActions.actions.push({
-          type: OPEN_TAB,
+          type: type,
           dynamicInfo: {
             name: action.file._text,
             location: actionToLocation(action),
--- a/src/ui/widgets/EmbeddedDisplay/bobParser.ts
+++ b/src/ui/widgets/EmbeddedDisplay/bobParser.ts
@@ -1,5 +1,5 @@
 import { REGISTERED_WIDGETS } from "../register";
-import { ComplexParserDict, parseWidget, ParserDict } from "./parser";
+import { ComplexParserDict, parseWidget, ParserDict, toArray } from "./parser";
 import {
   XmlDescription,
   OPI_COMPLEX_PARSERS,
@@ -7,7 +7,6 @@ import {
   OPI_PATCHERS,
   opiParseRules,
   opiParsePvName,
-  opiParseActions,
   opiParseColor,
   opiParseAlarmSensitive,
   opiParseString,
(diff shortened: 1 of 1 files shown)


src/ui/widgets/EmbeddedDisplay/bobParser.ts
```diff
@@ -217,6 +222,93 @@ function bobParseSymbols(jsonProp: ElementCompact): string[] {
...
+          openWebpageInfo: {
+            url: action.url._text,
+            description:
+              (action.description && action.description._text) || undefined
+          }
+        });
+      } else if (type === "open_display") {
+        const type = action.target._text === "replace" ? OPEN_PAGE : OPEN_TAB;
+        processedActions.actions.push({
+          type: type,
```
rjwills28, line 288: So I think the `type` sent here should always be `OPEN_TAB`, which is what the [opiParser](https://github.com/DiamondLightSource/cs-web-lib/blob/45773641ae3bdd93f224e3914d292509084b6312/src/ui/widgets/EmbeddedDisplay/opiParser.ts#L229) does. 

I think the action here opens another bob file in the same browser tab but can either be opened in the 'main' display or the 'details' display which are just arbitrary panels created on the page, i.e. 
Screenshot from 2025-02-04 15-46-04

But essentially, when I tested an action with the 'replace' type, which then sets the `type` to `OPEN_PAGE` I couldn't get anything to work, that is to say nothing happened. Only when I hardcoded the `type` to `OPEN_TAB` did the BOB file open in the 'main' panel (i.e. the panel on the left in the above screenshot). 

I guess `OPEN_PAGE` most only work from the json level then??

abigailalexander: It seems to all be quite confusing and took me a while to figure out. From what I can tell, those screens you're using in cs-web-proto are DynamicTab widgets, which means that using `OPEN_TAB` is the correct type. When I initially tested this in another project, I used a DynamicPage widget, which didn't work with `OPEN_TAB` but did with `OPEN_PAGE`. It seems like in order to decide whether `OPEN_TAB` or `OPEN_PAGE` should be used, some knowledge of the parent widget type is necessary. Having a `target` of "replace" doesn't seem to work in this DynamicTab widget, but "tab" does, and the opposite is true for a DynamicPage widget - a `target` of "replace" successfully replaces the current displayed screen, but "tab" does nothing.

If you want to test this, if you go to machineStatus.json in cs-web-proto and change one of the dynamicInfo actions to point towards a bob file with `location` "main" instead of opi, and then change the DynamicPage `location` at the bottom of the bob file to "main" as well, you should be able to open a bob file and then trigger a working replace action.

Maybe the best way to handle this is to document somewhere that "tab" screens need to be in DynamicTab widgets, and "replace" screens in DynamicPage widgets.

Regarding the "main" and "detail" locations - these seem like they're holdovers from when cs-web-proto first started, and like you said only really work when you create from json. I'm not sure what a better way to handle this is?

rjwills28: Thanks for pointing out how the DynamicPage works - I see how that is different now and I got your example working. It's weird how they depend on the json widget though. For example, take that machineStatus.json page in cs-web-proto. If I open a different OPI file that contains an action button to open another OPI file, that currently does not work because the underlying code in the opiParser returns the 'location' as 'main' whereas this json screen specifies the location as 'opi' and so there is no 'main' screen to display on hence nothing happens. So the code at the moment would require the client to always have a page with the location name 'main' and to display a tab it would have to be called 'details'. 

I guess it currently means that you can't have a BOB/OPI file that contains both actions to 'open in tab' and 'replace' as only one type would work depending on your 'master' json widget.

Also in terms of opening a 'tab' would we want it to open in a new browser tab? Or do we want it as a tab within the window? It's all pretty messy and confusing and if it's going to be used I think this concept of DynamicPages and DynamicTabs needs re-thinking. With all that said I think I'm going to approve the PR as it does fix the initial problem described. Then maybe we look at how to tidy this bit up later?

rjwills28 (approved):

src/ui/widgets/EmbeddedDisplay/bobParser.ts
abigailalexander: One other small change - I discovered that you can specify the name of which tab group to open a tab into, so I've updated that part accordingly so it defaults to "main" unless a name is given. This should allow you to open a new tab even if you only have one DynamicTab widget called "main", and if you want to open in another e.g. "details" you can specify it. This seems like a slightly more sensible way to handle new tabs in my opinion

I agree that the way it works now doesn't seem like it will hold up in future. Not being able to mix "replace" and "tab" action types isn't ideal, nor is necessitating a "main" location if you want to create screens dynamically. Having a discussion at some point on how to tidy this up will be useful. 

There are a few other things I've noticed to do with displaying files and the FileContext that I should to change. There are some small changes I will open a separate PR for, but also some bigger ones that I think will go hand in hand with the entire concept of DynamicPages and DynamicTabs and where/how we want them to open.

rjwills28: Sounds like a good plan - just let me know if there's anything I can do to help.

PR #76, opened 2025-02-06
Fix warnings in bobParser when alarmSensitive border property is missing

rjwills28:
While testing a previous change in the bobParser I noticed that I was getting warnings in the console that a property could not be parsed. This didn't cause errors in the application as the property was optional. 

Screenshot from 2025-02-06 09-16-32

The problem is that a BOB file generated from Phoebus by default does not include the `alarm_sensitive_border` property. By default this is assumed to be _true_ and so the property is only added if the user sets it to _false_. 

I've added a separate bobParser version to use instead of the opiParser version. Also updated the tests to make sure the default version is reflected correctly.

TESTING: To test with and without this fix, use a BOB file generated by Phoebus that contains a `TextUpdate` widget.

### DIFF
--- a/src/ui/widgets/EmbeddedDisplay/bobParser.test.ts
+++ b/src/ui/widgets/EmbeddedDisplay/bobParser.test.ts
@@ -107,7 +107,6 @@ describe("opi widget parser", (): void => {
       <y>62</y>
       <width>140</width>
       <height>50</height>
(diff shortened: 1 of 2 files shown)


abigailalexander (approved):
Tested and this fixes the issue mentioned, looks good to me.

PR #83, opened 2025-03-19
Fix image stretch to fit behaviour

NatLeung96:
I've added an objectPosition attribute, set to "top left", to the returned  component. I believe the default behaviour is for this to be set to "center" which would explain the behaviour. I don't if it would be better for there to be an option for this?

### DIFF
--- a/src/ui/widgets/Image/image.tsx
+++ b/src/ui/widgets/Image/image.tsx
@@ -38,17 +38,7 @@ export const ImageComponent = (
     }
   };
 
-  let imageHeight: string | undefined = undefined;
-  let imageWidth: string | undefined = undefined;
   const overflow = props.overflow ? "visible" : "hidden";
-  if (props.stretchToFit) {
-    imageWidth = "100%";
-    imageHeight = "100%";
-  } else if (props.fitToWidth) {
-    imageWidth = "100%";
-  } else if (props.fitToHeight) {
-    imageHeight = "100%";
-  }
 
   const style: CSSProperties = {
     ...commonCss(props as any),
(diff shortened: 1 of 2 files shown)


abigailalexander:
This looks good to me, I tested it in both Phoebus and CSStudio and the output image is aligned as we'd expect. 
Slightly separate to above, but I also noticed that the sizing of the Image widget is slightly wrong in the web UI compared to CSStudio if `stretchToFit=false`.
|Screenshot from 2025-03-19 15-22-48|Screenshot from 2025-03-19 15-22-32|
|:-:|:-:|
CSStudio | cs-web-lib

As you can see the CSStudio one expands to fill the full dimensions passed to the widget, so we probably aren't setting our dimensions quite right. It would be good to fix this in this PR as well.

NatLeung96:
I have now set the width and height of the div to always be "100%". Previously, they were only being set when stretchToFit was on.

src/ui/widgets/Image/image.tsx
```diff
@@ -54,8 +54,8 @@ export const ImageComponent = (
     ...commonCss(props as any),
     overflow,
     textAlign: "left",
-    width: imageWidth,
-    height: imageHeight
+    width: "100%",
```
abigailalexander, line 57: If we're setting the width and height here, we can probably remove the if-else statement on lines 44-51 that sets it based on `stretchToFit` and `fitToWidth`/`fitToHeight`, and the unused `imageHeight`/`imageWidth` variables.

abigailalexander (changes requested):
Looks correct now, thanks. Just one minor change to make

abigailalexander (approved):

PR #85, opened 2025-03-24
84 replace action button with mui button

NatLeung96:
I've changed the ActionButtonComponent so that it returns a MUI Button instead of a html button.

### DIFF
--- a/src/ui/widgets/ActionButton/actionButton.tsx
+++ b/src/ui/widgets/ActionButton/actionButton.tsx
@@ -18,7 +18,7 @@ import { Font } from "../../../types/font";
 import { Border } from "../../../types/border";
 import { MacroContext } from "../../../types/macros";
 import { ExitFileContext, FileContext } from "../../../misc/fileContext";
-import { Button, ThemeProvider} from "@mui/material";
+import { Button, ThemeProvider } from "@mui/material";
 
 import { defaultColours } from "../../../colourscheme";
 
@@ -78,10 +78,12 @@ export const ActionButtonWidget = (
           height: "100%",
           width: "100%",
           fontFamily: props.font?.css() ?? "",
-          color: props.foregroundColor?.toString() ??
-                     defaultColours.palette.primary.contrastText,
-          backgroundColor: props.backgroundColor?.toString() ??
-                     defaultColours.palette.primary.main,
+          color: 
+            props.foregroundColor?.toString() ??
+            defaultColours.palette.primary.contrastText,
+          backgroundColor: 
+            props.backgroundColor?.toString() ??
+            defaultColours.palette.primary.main,
           border: props.border?.css() ?? ""
         }}
         onClick={onClick}
--- a/src/ui/widgets/ActionButton/actionButton.tsx
+++ b/src/ui/widgets/ActionButton/actionButton.tsx
@@ -18,6 +18,9 @@ import { Font } from "../../../types/font";
 import { Border, BorderStyle } from "../../../types/border";
 import { MacroContext } from "../../../types/macros";
 import { ExitFileContext, FileContext } from "../../../misc/fileContext";
+import { Button } from "@mui/material";
+import { ThemeProvider } from "@mui/material";
+import { defaultColours } from "../../../colourscheme";
 
 export interface ActionButtonProps {
   text: string;
@@ -31,44 +34,6 @@ export interface ActionButtonProps {
   actions?: WidgetActions;
   visible?: boolean;
 }
-
-export const ActionButtonComponent = (
-  props: ActionButtonProps
-): JSX.Element => {
-  const style = commonCss(props);
-  if (props.disabled) {
-    style["cursor"] = "not-allowed";
-  }
-  style["whiteSpace"] = "normal";
-  // Use default button style if no border defined.
-  if (props.border?.style === BorderStyle.None) {
-    style["borderStyle"] = undefined;
-    style["borderWidth"] = undefined;
-    style["borderColor"] = undefined;
-    style["padding"] = "0";
-  }
-  return (
-    <button
-      className={classes.actionbutton}
-      disabled={props.disabled}
-      onClick={props.onClick}
-      style={style}
-    >
-      {props.image !== undefined ? (
-        <figure className={classes.figure}>
-          <img
-            style={{ width: "100%", display: "block" }}
-            src={props.image}
-            alt={props.image}
-          ></img>
-          <figcaption>{props.text}</figcaption>
-        </figure>
-      ) : (
-        <span style={{ whiteSpace: "normal" }}>{props.text || ""}</span>
-      )}
-    </button>
-  );
-};
 
 const ActionButtonPropType = {
   text: StringPropOpt,
@@ -104,18 +69,32 @@ export const ActionButtonWidget = (
       );
   }
   return (
-    <ActionButtonComponent
-      text={props.text ?? ""}
-      disabled={props.readonly}
-      onClick={onClick}
-      image={props.image}
-      backgroundColor={props.backgroundColor}
-      foregroundColor={props.foregroundColor}
-      font={props.font}
-      border={props.border}
-      actions={props.actions as WidgetActions}
-      visible={props.visible}
-    />
+    <ThemeProvider theme={defaultColours}>
+      <Button
+        variant="contained"
+        sx={{
+            height: "100%",
+            width: "100%",
+            fontFamily: props.font?.css() ?? "",
+            color: props.foregroundColor?.toString() ?? defaultColours.palette.primary.contrastText,
+            backgroundColor: props.backgroundColor?.toString() ?? defaultColours.palette.primary.main,
+            }}
+        onClick={onClick}
+      >
+      {props.image !== undefined ? (
+        <figure className={classes.figure}>
+          <img
+            style={{ width: "100%", display: "block" }}
+            src={props.image}
+            alt={props.image}
+          ></img>
+          <figcaption>{props.text}</figcaption>
+        </figure>
+      ) : (
+        props.text ?? ""
+      )}
+      </Button>
+    </ThemeProvider>
   );
 };
 
--- a/src/ui/widgets/ActionButton/actionButton.tsx
+++ b/src/ui/widgets/ActionButton/actionButton.tsx
@@ -29,8 +29,8 @@ export interface ActionButtonProps {
   enabled?: boolean;
   onClick: (event: React.MouseEvent<HTMLButtonElement>) => void;
   image?: string;
-  backgroundColor?: Color | string;
-  foregroundColor?: Color | string;
+  backgroundColor?: Color;
+  foregroundColor?: Color;
   border?: Border;
   font?: Font;
   actions?: WidgetActions;
@@ -57,7 +57,7 @@ export const ActionButtonComponent = (
 
   return (
     <Button
-      variant="contained"
+      variant={transparent ? "text" : "contained"}
       disabled={!enabled}
       fullWidth={true}
       sx={{
--- a/src/ui/widgets/ActionButton/actionButton.tsx
+++ b/src/ui/widgets/ActionButton/actionButton.tsx
@@ -20,7 +20,7 @@ import { Font } from "../../../types/font";
 import { Border } from "../../../types/border";
 import { MacroContext } from "../../../types/macros";
 import { ExitFileContext, FileContext } from "../../../misc/fileContext";
-import { Button, ThemeProvider } from "@mui/material";
+import { Button } from "@mui/material";
 
 import { diamondTheme } from "../../../diamondTheme";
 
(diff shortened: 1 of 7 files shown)


src/ui/widgets/ActionButton/actionButton.tsx
```diff
@@ -35,38 +39,39 @@ export interface ActionButtonProps {
...
+        sx={{
+          height: "100%",
+          width: "100%",
+          fontFamily: props.font?.css() ?? "",
+          color:
+            props.foregroundColor?.toString() ??
+            defaultColours.palette.primary.contrastText,
+          backgroundColor:
+            props.backgroundColor?.toString() ??
+            defaultColours.palette.primary.main,
```
abigailalexander, line 56: The default colours here don't match the default colours in Phoebus. For a button created in Phoebus with default settings, the button is grey with black text.

src/ui/widgets/ActionButton/actionButton.tsx
```diff
@@ -35,38 +39,39 @@ export interface ActionButtonProps {
 export const ActionButtonComponent = (
   props: ActionButtonProps
 ): JSX.Element => {
-  const style = commonCss(props);
-  if (props.disabled) {
-    style["cursor"] = "not-allowed";
-  }
```
abigailalexander, line 41: We want to keep this in as it provides a visual indication of when a button is disabled via the cursor in the same way Phoebus does.

src/ui/widgets/ActionButton/actionButton.tsx
```diff
@@ -35,38 +39,39 @@ export interface ActionButtonProps {
 export const ActionButtonComponent = (
   props: ActionButtonProps
 ): JSX.Element => {
-  const style = commonCss(props);
```
abigailalexander, line 38: Removing this `commonCss` function also removes the ability to set the button `backgroundColour` as transparent and some other CSS. It would be good to either keep using this function, or ensure we're still setting all the needed CSS.

src/ui/widgets/ActionButton/actionButton.tsx
```diff
@@ -35,38 +39,39 @@ export interface ActionButtonProps {
...
-  // Use default button style if no border defined.
-  if (props.border?.style === BorderStyle.None) {
-    style["borderStyle"] = undefined;
-    style["borderWidth"] = undefined;
-    style["borderColor"] = undefined;
-    style["padding"] = "0";
-  }
   return (
-    <button
-      className={classes.actionbutton}
```
abigailalexander, line 52: If the text for the button is too long, it now overflows its bounds. This was previously handled by the CSS class used here that has been removed. We want to either keep using this CSS class (if it doesn't interfere with MUIs own CSS handling) or move some of the properties set in the CSS file over to set them here in the component.

src/ui/widgets/ActionButton/actionButton.tsx
```diff
@@ -35,38 +39,39 @@ export interface ActionButtonProps {
...
-            src={props.image}
-            alt={props.image}
-          ></img>
-          <figcaption>{props.text}</figcaption>
-        </figure>
-      ) : (
-        <span style={{ whiteSpace: "normal" }}>{props.text || ""}</span>
-      )}
-    </button>
+    <ThemeProvider theme={defaultColours}>
```
abigailalexander, line 43: Typically an MUI theme is defined at a top level in the code, and that flows down into all child components. Rather than specifying this here, I'd suggest moving this definition into something like the Display component, so it can be automatically applied to everything beneath it.

src/ui/widgets/ActionButton/actionButton.tsx
```diff
@@ -35,38 +39,39 @@ export interface ActionButtonProps {
...
+      <Button
+        variant="contained"
+        disabled={props.disabled}
+        sx={{
+          height: "100%",
+          width: "100%",
+          fontFamily: props.font?.css() ?? "",
```
abigailalexander, line 50: If no font is provided, we probably want to set a default one that matches Phoebus (aka Liberation Sans size 14). Specifying this in the theme alongside default colours would be ideal.

Additionally, rather than checking here whether a property is undefined or not, it's a little neater to do that at the top of the widget and set defaults there. If you look at BoolButton, you should be able to see what I mean e.g.

```
  const {
   ...
    backgroundColor = defaultColours.palette.primary.main, // Set your default value here
    foregroundColor = ...,
   ...
  } = props;
```
This unpacks the props into variables, and sets the value passed in on the right if the property is undefined

src/ui/widgets/ActionButton/actionButton.tsx
```diff
@@ -35,38 +39,39 @@ export interface ActionButtonProps {
...
-          <figcaption>{props.text}</figcaption>
-        </figure>
-      ) : (
-        <span style={{ whiteSpace: "normal" }}>{props.text || ""}</span>
-      )}
-    </button>
+    <ThemeProvider theme={defaultColours}>
+      <Button
+        variant="contained"
+        disabled={props.disabled}
```
abigailalexander, line 46: This doesn't seem to be setting properly, it looks like the prop used in the .bob/.opi file is `enabled` instead of `readonly` property that is then mapped to `disabled` later on line 116. (It appears this is an existing bug in this code, but it would be good to fix it here)

src/ui/widgets/ActionButton/actionButton.tsx
```diff
@@ -11,13 +11,17 @@ import {
...
 } from "../propTypes";
 import { Color } from "../../../types/color";
 import { Font } from "../../../types/font";
-import { Border, BorderStyle } from "../../../types/border";
+import { Border } from "../../../types/border";
 import { MacroContext } from "../../../types/macros";
 import { ExitFileContext, FileContext } from "../../../misc/fileContext";
+import { Button, ThemeProvider } from "@mui/material";
+
+import { defaultColours } from "../../../colourscheme";
```
abigailalexander, line 24: This is a little bit of a nitpick, but file names should follow camelcase naming convention, and `colourscheme.ts` should be `colourScheme.ts`.

abigailalexander (changes requested):
Overall this looks good, but there are a couple of cases where the behaviour isn't as expected. I've mentioned most in comments, but to summarise the big ones:

- The default colours used in your theme are different to Phoebus
- Text overflow and wrapping isn't handled
- If you set your button size down to 20x20 pixels, you'll notice it doesn't shrink small enough. This is because MUI has a `minWidth` and `minHeight` property, and also applies padding on top of the width you've specified (you can see this also by inspecting the component and comparing the size of the top level div with the MUI button). Our 20x20 component is now 64x24 in size.

I attached some pictures below to show the difference. The left one is how it should look. 
(You'll notice the font also doesn't rotate on the transparent button. This is an existing bug, so if you want you can fix it in this PR, or we'll open an issue for it to be fixed at a separate time)

|Screenshot from 2025-03-28 09-40-20|Screenshot from 2025-03-28 09-40-11|
|:-:|:-:|

src/ui/widgets/ActionButton/actionButton.tsx
NatLeung96: I've changed the theme name from defaultColours to diamondTheme and change the default colours to match Phoebus defaults.

src/ui/widgets/ActionButton/actionButton.tsx
NatLeung96: Managed to get text overflow behaviour to somewhat resemble that of Phoebus.

src/ui/widgets/ActionButton/actionButton.tsx
NatLeung96: Moved ThemeProvider to embeddedDisplay.

src/ui/widgets/ActionButton/actionButton.tsx
NatLeung96: Set a default font in typography object of diamondTheme.

src/ui/widgets/ActionButton/actionButton.tsx
NatLeung96: Renamed colourscheme.ts to diamondTheme.ts and renamed defaultColours to diamondTheme.

src/ui/widgets/ActionButton/actionButton.tsx
NatLeung96: I've made it so that the transparent field from the .bob file actually gets passed to the widget and added a test to make the background transparent if it is set to true. The way it currently works is that, when transparent, the button colour matches the background but it keeps the button shadow. Personally, I think this looks cool but I can remove it if desired.

NatLeung96:
There is still an issue with the text rotation. I'm not 100% sure if it is a parser problem or a case of the rotationStep variable not being passed to the widget properly. I tried adding a "rotation_step" token to the simple parser dictionary but, in testing, it returns undefined instead of a number.

Regardless, I've added an implementation that should rotate the text on the assumption that a rotationStep is passed.

abigailalexander (changes requested):
Overall looks a lot closer to Phoebus now, thanks for doing those changes! I noticed the the very small button (20x20 px) still doesn't size right and I think this is due to the `lineHeight` property. This is scaled by the font and dictates the minimum height of a component based on the height of the text. I'd try setting this property to `"unset"` and check if it works fine

Also, this isn't a necessary change but might be good for code organisation: MUI has a [styled](https://mui.com/system/styled/?srsltid=AfmBOooBAjMk7dh32UvaW3nAgLXIuQNTToewq3zoaTT6ADaXVADqbo40) functionality that is useful for static style modification as we can redefine the style of the base component. It just makes everything a bit tidier and easier to read, and I'd suggest using it for things like `"&.MuiButton-root"` properties that are static and not affected by the props. e.g.

```
// Rename the import so we don't end up duplicate component names after styling
import { Button as MuiButton } from "@mui/material";

const Button = styled(MuiButton)({
   // Put all the static styling here
    display: "block",
    "&.MuiButton-root": {
        minWidth: 0
        etc...
    }
}

export const ActionButtonComponent = (
  props: ActionButtonProps
): JSX.Element => {
  // All the code here
   ...
   return (
     <Button //This has the static styling set, now we define dynamic styling
        sx={color: foregroundColor.toString(), ...}
     > 
     ...
  }
```

Not every component will need dedicated styling, but if we're setting a lot of properties it might be good to use.

Lastly, for `rotationStep` to be passed as a prop it needs to be added to the parser list. This can be found in `src/ui/widgets/EmbeddedDisplay/bobParser.ts` in the `parseBob` function. If you add it to the list of properties there, it should hopefully work :)

src/ui/widgets/ActionButton/actionButton.tsx
```diff
@@ -11,48 +11,79 @@ import {
...
 export interface ActionButtonProps {
   text: string;
-  disabled?: boolean;
+  enabled?: boolean;
   onClick: (event: React.MouseEvent<HTMLButtonElement>) => void;
   image?: string;
-  backgroundColor?: Color;
-  foregroundColor?: Color;
+  backgroundColor?: Color | string;
+  foregroundColor?: Color | string;
```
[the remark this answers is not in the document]
NatLeung96: Nope, you're right. That was me over correcting for a type clash between Phoebus output and Theme.

src/ui/widgets/ActionButton/actionButton.tsx
```diff
@@ -64,9 +95,18 @@ export const ActionButtonComponent = (
           <figcaption>{props.text}</figcaption>
         </figure>
       ) : (
-        <span style={{ whiteSpace: "normal" }}>{props.text || ""}</span>
+        <span
+          style={{
+            display: "block",
+            transform: `rotation(${rotationStep * -90}deg)`.toString(),
+            whiteSpace: "nowrap",
+            textOverflow: "ellipsis"
```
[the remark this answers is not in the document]
NatLeung96: I will defer to your judgement on this matter. Trying to figure this out was a pain.

NatLeung96:
> Overall looks a lot closer to Phoebus now, thanks for doing those changes! I noticed the the very small button (20x20 px) still doesn't size right and I think this is due to the `lineHeight` property. This is scaled by the font and dictates the minimum height of a component based on the height of the text. I'd try setting this property to `"unset"` and check if it works fine

"unset" doesn't seem to do anything. Apparently, the default behaviour is dependent on the renderer. Instead, I set it manually to 1 (i.e. equal to the height of the font). 

> Also, this isn't a necessary change but might be good for code organisation: MUI has a [styled](https://mui.com/system/styled/?srsltid=AfmBOooBAjMk7dh32UvaW3nAgLXIuQNTToewq3zoaTT6ADaXVADqbo40) functionality that is useful for static style modification as we can redefine the style of the base component. It just makes everything a bit tidier and easier to read, and I'd suggest using it for things like `"&.MuiButton-root"` properties that are static and not affected by the props. e.g.

> Lastly, for `rotationStep` to be passed as a prop it needs to be added to the parser list. This can be found in `src/ui/widgets/EmbeddedDisplay/bobParser.ts` in the `parseBob` function. If you add it to the list of properties there, it should hopefully work :)

abigailalexander (approved):
Looks good to me!

PR #89, opened 2025-03-26
Replaced Checkbox with MUI Checkbox

NatLeung96:


src/ui/widgets/Checkbox/checkbox.tsx
```diff
@@ -31,33 +37,37 @@ export type CheckboxComponentProps = InferWidgetProps<typeof CheckboxProps> &
...
+        sx={{
+          color:
+            props.foregroundColor?.toString() ??
+            defaultColours.palette.primary.contrastText,
+          "&:npt($checked) .MuiIconButton-label:after": {
+            color:
+              props.foregroundColor?.toString() ??
+              defaultColours.palette.primary.main
+          }
+        }}
```
abigailalexander, line 58: A few notes here:

- The theme colours don't match Phoebus defaults, as mentioned in the other PR as well
- For the label, it would be good to have word wrapping and font settings (This is something that was not included in the old version but should have been, so would be good to add in)

src/ui/widgets/Checkbox/checkbox.tsx
```diff
@@ -31,33 +37,37 @@ export type CheckboxComponentProps = InferWidgetProps<typeof CheckboxProps> &
 export const CheckboxComponent = (
   props: CheckboxComponentProps
 ): JSX.Element => {
-  const style = {
-    ...commonCss(props as any),
-    display: "flex",
-    alignItems: "center",
-    cursor: "pointer"
-  };
```
abigailalexander, line 39: I think these style settings are worth keeping and applying to the overall component.

abigailalexander (changes requested):
Looks good overall, I only had a few small comments that are quite similar to those on the ActionButton PR.

I did notice that the Checkbox doesn't seem to be located correctly in the page, it's setting some padding that makes it shift down and to the right of where it should be compared to Phoebus (see images - left is the size and position of the widget as defined in Phoebus, right shows how the padding is forcing the entire component including the label to be shifted)
|Screenshot from 2025-03-31 08-50-38|Screenshot from 2025-03-31 08-50-47|
|:-:|:-:|

abigailalexander (commented):
Thanks for fixing those other items. Just the issue with the padding/sizing of the checkbox that I mentioned to you remaining. From what I saw when I looked into it:

- Unneeded padding on the MUI Checkbox component itself
- Unnecessary right-side margin on the MUI Form Control element
- Some sort of size restriction that doesn't allow the Checkbox Icon to go small enough. It seems like it might be one of the MUI internal components that needs adjusting e.g. `MuiCheckbox-root`

PR #92, opened 2025-04-03
86 replace boolean button with mui button

NatLeung96:


src/ui/widgets/BoolButton/boolButton.test.tsx
```diff
@@ -135,11 +135,11 @@ describe("<BoolButton />", (): void => {
     const button = getByRole("button") as HTMLButtonElement;
 
     // Original on values
-    expect(button.style.backgroundColor).toEqual("rgb(0, 235, 10)");
+    // expect(button.style.backgroundColor).toEqual("rgb(0, 235, 10)");
```
abigailalexander, line 138: If these tests are failing because of the MUI issue, I think some snapshot tests should be created to cover the same use case (testing background/led colour). Currently this final test is defunct and isn't actually testing anything at all as the assertions are commented out, so it's "passing" only because it isn't checking anything.

It looks like it should be possible to [use snapshots alongside an onClick event for this final test](https://stackoverflow.com/questions/56021640/react-testing-library-taking-a-snapshot-after-fireevent). I'd suggest having 3 snapshots: one for default values, one for overriding defaults, one for triggering a click.

src/ui/widgets/BoolButton/boolButton.test.tsx
```diff
@@ -35,10 +35,10 @@ describe("<BoolButton />", (): void => {
...
 
     expect(button.textContent).toEqual("On");
-    expect(button.style.height).toEqual("30px");
-    expect(button.style.width).toEqual("100px");
-    expect(button.style.backgroundColor).toEqual("rgb(200, 200, 200)");
-    expect(button.style.borderRadius).toEqual("");
+    // expect(button.style.height).toEqual("30px");
+    // expect(button.style.width).toEqual("100px");
+    // expect(button.style.backgroundColor).toEqual("rgb(200, 200, 200)");
+    // expect(button.style.borderRadius).toEqual("");
```
abigailalexander, line 41: These should be deleted rather than just commented out, as with the others.

abigailalexander (changes requested):
Appearance looks great, just some notes about testing the style.

abigailalexander (approved):
All good, tests are doing exactly what we'd expect :)

PR #93, opened 2025-04-03
90 replace label with mui typography

NatLeung96:


src/ui/widgets/Label/label.tsx
```diff
@@ -41,61 +42,57 @@ export const LabelComponent = (
...
-  }
-
-  if (wrapWords) {
-    style["wordBreak"] = "break-word";
-    style["whiteSpace"] = "break-spaces";
-  }
 
   // Simple component to display text - defaults to black text and dark grey background
   return (
-    <div className={className} style={style}>
```
abigailalexander, line 96: I'd suggest either:

- Keeping using this CSS class. It already contains properties like `width: "100%"`, `height: "100%"` and display: "flex" which saves us duplicating those lines a little below this.
- Don't use the CSS class, but delete the `label.module.css `file as it is now obsolete.

src/ui/widgets/Label/label.test.tsx
```diff
@@ -13,16 +13,16 @@ describe("<Label />", (): void => {
...
+  //   const label = screen.getByText("hello");
+  //   // The text is in the span, not the parent div.
+  //   if (label.parentElement && "style" in label.parentElement) {
+  //     expect(label.parentElement.style).toHaveProperty(
+  //       "backgroundColor",
+  //       "transparent"
+  //     );
+  //   }
+  //   expect.assertions(1);
+  // });
```
abigailalexander, line 27: If this test has been commented out because it doesn't work with MUI (as you mentioned), I'd suggest adding another snapshot to cover the same use case (transparent vs colour background). I'd also delete rather than just comment this out.

src/ui/widgets/Label/label.tsx
```diff
@@ -41,61 +42,57 @@ export const LabelComponent = (
...
-  const style: CSSProperties = commonCss(editedProps);
   const {
+    transparent = true,
+    foregroundColor = diamondTheme.palette.primary.contrastText,
     textAlign = "left",
     textAlignV = "top",
     text = "",
-    rotationAngle,
+    rotationStep = 0,
     wrapWords
```
abigailalexander, line 52: This is an extra point, but `wrapWords` defaults to true in Phoebus. It would be good to add this in here

abigailalexander (changes requested):
Nice, looks really good! I think the only issue visually is the text overflow. Phoebus hides overflow text, but it's showing here.

abigailalexander (approved):
All good now

PR #99, opened 2025-04-07
91 replace input with mui textfield

NatLeung96:


src/ui/widgets/Input/input.tsx
```diff
@@ -32,73 +33,149 @@ export interface InputProps {
...
+    borderColor: "#000000"
+  },
+  "&:hover .MuiOutlinedInput-root .MuiOutlinedInput-notchedOutline": {
+    border: "1",
+    borderColor: "#000000"
+  }
+});
+
 export const SmartInputComponent = (
   props: PVInputComponent & {
```
abigailalexander, line 74: I think this was already being done an odd way, but it's worth changing the way the props are handled so it's similar to  Readback or Label e.g.

```
export type InputComponentProps = InferWidgetProps<typeof InputProps> &
  PVComponent;

// inside Component definition
props: InputComponentProps
```

We also seem to have 2 prop definitions - InputProps and InputWidgetProps. I think InputWidgetProps is the one to keep, just ensure all the properties we need are in that list (I don't think we need the `on...` events)

abigailalexander (changes requested):
Looks good! Most of the issues I'm seeing seem to be existing issues not ones related to your changes, such as:

- Multi-line and text wrapping
- In Phoebus you can set a border, it would be good to include this

It appears we were doing things quite weirdly in this widget, so your changes are much appreciated. Also don't forget the CSS file ;)

NatLeung96:
This has been a wild revision.

1. Borders
After playing around and testing things out, I finally realised that, once the border thickness has been set, it cannot be changed. &:hover can only change the colour and &.Mui-focused changes the colour with the border thickness baked-in at 2px. The result is that different buttons will have different behaviours depending on their borders. For example, an input widget without a border won't look different when the mouse is hovered over it. BTW, the default thickness is 0px and this is coded into the parser itself.

2 PV props
I'm hoping that I got the props type declarations properly set up like the other widgets. This did highlight that there may have been a bug with the definition of PVInputComponent type as pvName was declared twice with different props (one with PvPropOpt - i.e. PV - and the other with string). I took the string definition to be incorrect and changed that to PV. As a result, I had to change some variables and edit some tests. Hopefully that was the correct thing to do.

3. Blur
I also learnt that it is currently not possible to trigger blur on MUI TextField. With the old widget, after typing a value and hitting Enter, the value would get written to the pv and the widget would become unfocussed. Now, it just writes the pv without unfocussing the widget, which could potentially be dangerous.

abigailalexander (changes requested):
The appearance looks really good now! Just the two comments below

PR #102, opened 2025-04-15
96 replace slidercontrol with mui slider

NatLeung96:


NatLeung96:
Functionally, it should hopefully work all fine. Aesthetically, it may need a bit of work.

src/ui/widgets/SlideControl/slideControl.tsx
```diff
@@ -1,115 +1,177 @@
...
+  transparent: BoolPropOpt,
+  levelHihi: FloatPropOpt,
+  levelHigh: FloatPropOpt,
+  levelLow: FloatPropOpt,
+  levelLolo: FloatPropOpt,
+  showHihi: BoolPropOpt,
+  showHigh: BoolPropOpt,
+  showLow: BoolPropOpt,
+  showLolo: BoolPropOpt,
+  increment: FloatPropOpt
```
abigailalexander, line 45: We're missing a `showScale` property here (which turns on or off the scale labels)

src/ui/widgets/SlideControl/slideControl.tsx
```diff
@@ -1,115 +1,177 @@
...
+      orientation={horizontal ? "horizontal" : "vertical"}
+      onChange={(_, newValue) => setInputValue(newValue as number)}
+      onChangeCommitted={(_, newValue) => onMouseUp(newValue as number)}
+      valueLabelDisplay="auto"
+      min={min}
+      max={max}
+      marks={marks}
+      step={increment}
+      sx={{
+        color: backgroundColor.toString(),
```
abigailalexander, line 146: This widget seems a bit odd in Phoebus and sets the colour of the slider based off the `foregroundColor` rather than `backgroundColor`, which I think we should keep doing. Having the entire slider be a single colour looks better than just the thumb part.

src/ui/widgets/SlideControl/slideControl.tsx
```diff
@@ -1,115 +1,177 @@
...
+        "& .MuiSlider-thumb": {
+          height: 24,
+          width: 24,
```
abigailalexander, line 149: 24x24 is quite large, something like 16x16 might work better here.

abigailalexander (changes requested):
Overall I think it looks good aesthetically, and it's functioning exactly as it should be. Only a few minor tweaks but otherwise good!

abigailalexander (approved):
Looks good now!

PR #103, opened 2025-04-22
94 replace readback with mui textfield

NatLeung96:
I did a lot of calculations for this one, particularly for the font size. There's good chance that I may have overcomplicated this a bit so please let me know if you have any suggestions for possibly simplifying it.


src/ui/widgets/Readback/readback.tsx
```diff
@@ -108,42 +124,115 @@ export const ReadbackComponent = (
...
+          lineHeight: 1
+        },
+        "& .MuiInputBase-root": {
+          alignItems: alignmentV,
+          color: foregroundColor,
+          backgroundColor: backgroundColor
+        },
+        "& .MuiOutlinedInput-root": {
+          "& .MuiOutlinedInput-notchedOutline": {
+            border: border,
```
abigailalexander, line 220: I'm noticing that when there's a border present that's more than a few pixels thick it seems to cut off some of the text inside the box. This seems like an MUI feature but it does mean it's hard to read some text displayed (in comparison, Phoebus has the border expand outside the width and height of the component). It would be good to explore if this can be fixed.
Screenshot from 2025-04-23 15-44-31

abigailalexander (commented):
Looks really good overall! Just a slight overlap in border/text that would be good to see if it can be fixed, or if there's a way around it.

PR #105, opened 2025-04-30
Make input use same border stying as readback

NatLeung96:
Changed Input widget to use same border styling as Readback widget

### DIFF
--- a/src/ui/widgets/Input/input.tsx
+++ b/src/ui/widgets/Input/input.tsx
@@ -11,11 +11,13 @@ import {
   ColorPropOpt,
   BoolPropOpt,
   BorderPropOpt,
-  StringPropOpt
+  StringPropOpt,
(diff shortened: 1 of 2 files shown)


src/ui/widgets/Input/input.tsx
```diff
@@ -130,7 +172,7 @@ export const SmartInputComponent = (
     <TextField
       disabled={!enabled}
       value={inputValue}
-      multiline={multiLine}
```
abigailalexander, line 133: The `multiLine` property should still be set here (alongside `maxRows`) otherwise we lose the ability to write across multiple lines.

abigailalexander (changes requested):
The border looks good now and isn't taking up space inside the text box, just some minor tweaks that were affected by these changes.

abigailalexander (approved):

PR #115, opened 2025-05-07
Update widget rules not applying to nested properties

NatLeung96:
Added x and y fields to the RelativePosition prop as Widgets expects a PositionProp, which can either be RelativePosition or AbsolutePosition, and therefore it wasn't possible to change the x and y fields of the PositionProp because they may not always exist.

### DIFF
--- a/src/types/position.ts
+++ b/src/types/position.ts
@@ -70,6 +70,8 @@ export class AbsolutePosition {
 }
 
 export class RelativePosition {
+  public x: string;
+  public y: string;
   public width: string;
   public height: string;
   public margin: string;
@@ -79,6 +81,8 @@ export class RelativePosition {
   public minHeight: string;
 
   public constructor(
+    x = "",
+    y = "",
     width = "",
     height = "",
     margin = "",
@@ -87,6 +91,8 @@ export class RelativePosition {
     maxWidth = "",
     minHeight = ""
   ) {
+    this.x = x;
+    this.y = y;
     this.width = width;
     this.height = height;
     this.margin = margin;
--- a/src/ui/hooks/useRules.ts
+++ b/src/ui/hooks/useRules.ts
@@ -10,6 +10,7 @@ import { AlarmQuality, DType } from "../../types/dtypes";
 import { SubscriptionType } from "../../connection/plugin";
 import { Border, BorderStyle } from "../../types/border";
 import { Color } from "../../types/color";
+import { opiParseColor } from "../widgets/EmbeddedDisplay/opiParser";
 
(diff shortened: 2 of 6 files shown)


abigailalexander (commented):
It looks like it's working to me, but I've only had a chance to take a quick look over it so I'm adding Becky as a reviewer for a second pair of eyes.

Also, would it be possible for you to add a short explanation on this PR about how adding x,y coordinates fixes the position rules? Is it just that you can't set the properties`x`, `y` if they don't always exist?

rjwills28:
Hi @NatLeung96 - I'm having a few problems getting screens to display properly with this branch. Can you let me know if I've missed anything or is you see the same thing? I'll can have a closer look at what's going on but as I said, want to make sure it's not just me.

- If I run the **machine-status** app against this version of cs-web-lib and set the `VITE_BUILD_TARGET=screen` in the .env file I am getting:

  Screenshot from 2025-05-09 10-02-53

  Note: the `VITE_BUILD_TARGET=public ` screen looks OK.

- I get similar strange displays when I also test against **cs-web-proto**:

  Old version:
  Screenshot from 2025-05-09 10-07-20

  What it looks like now:
  Screenshot from 2025-05-09 10-04-44

rjwills28:
Hi @NatLeung96 - thanks for fixing that issue -  I can confirm that it is now fixed.

I was able to test this against a BOB screen I created where I had attached different rules to widgets associated with changing the x, y positions and border colour & width. These all work well and in the same way as they do in Phoebus.

However, I've just seen a few more problems in general with some of the displays in **cs-web-proto** (another app that uses cs-web-lib).

This is what the display should look like:
Screenshot from 2025-05-12 10-36-08

But this is now what is being rendered:
Screenshot from 2025-05-12 11-06-08
^ As you can see, the 'Open screens' button in the very top left is mis-aligned and for some reason this screen is not showing. I think this is a tabbed pane.

Could you try running this branch up against cs-web-proto? Available: https://gitlab.diamond.ac.uk/controls/web/cs-web-proto.  Then go to 'Open screens' > 'BL21I'.

NatLeung96:
That's because I also forgot to add x and y props to the displayWidget position in the opiParser. That has fixed the hardware tabs but the "Open screens" button is still fudged but that appears to be due to code in cs-web-proto. In the header.tsx file, a DrawerWidget is created with a RelativePosition that now no longer has enough input arguments.

rjwills28 (approved):
Great! Thank you for fixing that and pointing out the specific issue in cs-web-proto. All now looks good to me.

PR #116, opened 2025-05-13
81 pv formulae not supported

NatLeung96:
PV parser has been updated to correctly parse PV formulae and to parse PV starting with "=" to start with "eq://". "eq://" has been added to list of plugins in store.ts to enable formulae subscribe requests.

### DIFF
--- a/src/types/pv.ts
+++ b/src/types/pv.ts
@@ -24,9 +24,16 @@ export class PV {
    * PV.parse("pvName")
    */
   public static parse(pvName: string, defaultProtocol = "ca"): PV {
+    if (pvName.startsWith("=")) {
+      return new PV(pvName.slice(1), "eq");
+    }
     if (pvName.includes(PV.DELIMITER)) {
-      const parts = pvName.split(PV.DELIMITER);
-      return new PV(parts[1], parts[0]);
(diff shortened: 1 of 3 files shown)


src/types/pv.ts
```diff
@@ -41,7 +48,8 @@ export class PV {
   public qualifiedName(): string {
     // This can happen if the name is substituted by a macro
     // after the PV object has been created.
-    if (this.name.includes(PV.DELIMITER)) {
+    // Need to make sure that the PV.DELIMITER is not associated with a nested PV.
+    if (this.name.includes(PV.DELIMITER) && !this.name.includes("`")) {
```
rjwills28, line 52: Very minor point - it looks like you can also create a formula using a single quote (i.e. `'`) as well as with a backtick, e.g. =3+4*'sim://ramp'. This works in Phoebus and PVWS so maybe we could just add an addition `&&` to check that it also doesn't include a `'`?

rjwills28 (commented):
This look really good! I've tested it out and all works really nicely. I just noticed one very minor point that could be addressed before merging - see inline comment.

NatLeung96:
Also made that else-if check for "=".

rjwills28 (approved):
Looks great - thanks for making those changes.

PR #117, opened 2025-05-19
Generalise use of themes

NatLeung96:
Currently, the widgets have hard-coded default styling. In this branch, these default styles have been replaced with theme variables that the user can provide using MUI's ThemeProvider. If the user doesn't provide a theme, it will just use MUI's default theme. One issue is that the EmbeddedDisplay widget is unable to read the theme provided by ThemeProvider and therefore all of its child widgets won't inherit the theme. To circumvent this issue, a theme prop has been added to the EmbeddedDisplay widget which it can re-inject using a ThemeProvider wrapped around its child widgets.


src/ui/widgets/EmbeddedDisplay/embeddedDisplay.tsx
```diff
@@ -209,17 +216,21 @@ export const EmbeddedDisplay = (
 
   if (props.border?.style === BorderStyle.GroupBox) {
     return (
-      <MacroContext.Provider value={embeddedDisplayMacroContext}>
-        <GroupBoxComponent name={resolvedName} styleOpt={0}>
-          <ThemeProvider theme={diamondTheme}>{component}</ThemeProvider>
-        </GroupBoxComponent>
-      </MacroContext.Provider>
+      <ThemeProvider theme={props.theme ?? theme}>
```
abigailalexander, line 219: Rather than using the default MUI theme here if the client application doesn't pass one in, we should use a "default" theme we have created that has Phoebus defaults. So the diamondTheme you previously created (which contains Phoebus defaults for font and colours etc) could be used (but perhaps renamed to `phoebusTheme` or `defaultTheme`).

abigailalexander (changes requested):
Looks mostly good to me, two things I noted:

- As I mentioned in a comment, we probably want a default Phoebus theme rather than the default MUI theme
- Currently, I can't seem to pass a theme to EmbeddedDisplay in Daedalus to overwrite the default. It gives this error: `Property 'containerQueries' is missing in type 'import("daedalus/node_modules/@mui/material/styles/createTheme").Theme' but required in type 'import("/cs-web-lib/node_modules/@mui/material/styles/createThemeNoVars").Theme'.ts(2741)
cssContainerQueries.d.ts(10, 5): 'containerQueries' is declared here.`

Have you tested passing a custom theme to EmbeddedDisplay and did it work for you?

abigailalexander:
The issue with passing the theme from a client was on my end, that all works fine :)

abigailalexander (approved):

PR #119, opened 2025-05-20
Update label borders and rotation

NatLeung96:
This branch adds rounded borders to the label widget and changes the border styling to match Phoebus behaviour. In addition, the translation of the widget after rotating has been fixed. Currently, the widget appears correct when it is rotated 0, 90, and 180 degrees but will appear in the wrong place when rotated -90/270 degrees.

Sorry, the rebase was not as straightforward as I would have liked.


abigailalexander (approved):
Looks good to me!

PR #120, opened 2025-05-21
Update rotation of actionbutton and readback

NatLeung96:
This branch fixes the bug with the readback widget where it appears in the wrong place when rotated -90 degrees. The rotation of the actionbutton widget has also been changed to match the rotation behaviour of the label and readback widgets.

### DIFF
--- a/src/ui/widgets/ActionButton/actionButton.tsx
+++ b/src/ui/widgets/ActionButton/actionButton.tsx
@@ -17,9 +17,7 @@ import {
 } from "../propTypes";
 import { MacroContext } from "../../../types/macros";
 import { ExitFileContext, FileContext } from "../../../misc/fileContext";
-import { styled, Button as MuiButton } from "@mui/material";
-
-import { diamondTheme } from "../../../diamondTheme";
+import { styled, Button as MuiButton, useTheme } from "@mui/material";
 import { WIDGET_DEFAULT_SIZES } from "../EmbeddedDisplay/bobParser";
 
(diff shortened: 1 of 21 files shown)


src/ui/widgets/ActionButton/actionButton.tsx
```diff
@@ -75,16 +79,39 @@ export const ActionButtonComponent = (
...
+        return `rotate(${rotationStep * -90}deg)`;
+      case 1: // 90 degrees
+        return `rotate(${rotationStep * -90}deg) translateY(${offset}px) translateX(${offset}px)`;
+      case 3: // -90 degrees
+        return `rotate(${rotationStep * -90}deg) translateY(${-offset}px) translateX(${-offset}px)`;
+      default: // Unreachable
+        return "";
+    }
+  })();
+
```
abigailalexander, line 99: I think it would be best to move this section calculating the rotation into a separate function inside the utils.ts file, so it can be reused across widgets (like the Label widget) and prevents code duplication.

abigailalexander (changes requested):
All works fine, just one suggestion regarding moving the rotation calculation into a separate function that can be reused elsewhere.

src/ui/widgets/ActionButton/actionButton.tsx
NatLeung96: Good idea

abigailalexander (approved):
Nice, thanks!

PR #123, opened 2025-09-16
Add Tank widget using mui x-charts

GregJHarris:
This change adds the Tank widget by making use of the MUI X-Charts BarChart component, #118 
* New import of @mui/x-charts

Side changes:
* The MUI package @mui/material  was upgraded from 6.1.6 to 7.3.2, the more recent version provides better support for ESM imports that are required to resolve some X-Charts dependencies. 
* Snapshot update following on from the MUI upgrade.
* In certain cases wrapping `{props.children}`  in fragments to become `<>{props.children}</>` to resolve `Type 'bigint' is not assignable to type 'ReactNode'` type error

Screenshots:
* Vertical logscale with tooltip showing.
image

* Horizontal linear scale:
image

### DIFF
--- a/src/ui/widgets/EmbeddedDisplay/opiParser.ts
+++ b/src/ui/widgets/EmbeddedDisplay/opiParser.ts
@@ -637,7 +637,7 @@ export const OPI_SIMPLE_PARSERS: ParserDict = {
   textAlignV: ["vertical_alignment", opiParseVerticalAlignment],
   backgroundColor: ["background_color", opiParseColor],
   foregroundColor: ["foreground_color", opiParseColor],
-  emptyColor: ["empty_color", opiParseColor],
+  emptyColor: ["color_fillbackground", opiParseColor],
   onColor: ["on_color", opiParseColor],
   offColor: ["off_color", opiParseColor],
   fillColor: ["fill_color", opiParseColor],
--- a/src/ui/widgets/Tank/tank.tsx
+++ b/src/ui/widgets/Tank/tank.tsx
@@ -44,8 +44,8 @@ export const TankComponent = (
     showLabel = false,
     font,
     horizontal = false,
-    fillColor = Color.fromRgba(0, 0, 255, 255),
-    emptyColor = Color.fromRgba(250, 250, 250, 255),
+    fillColor = Color.fromRgba(0, 0, 255, 1),
+    emptyColor = Color.fromRgba(192, 192, 192, 1),
     precision = undefined,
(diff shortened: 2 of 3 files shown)


src/ui/widgets/Tank/tank.tsx
```diff
@@ -0,0 +1,184 @@
...
+): JSX.Element => {
+  const {
+    value,
+    pvName,
+    limitsFromPv = false,
+    showLabel = false,
+    font,
+    horizontal = false,
+    fillColor = Color.fromRgba(0, 0, 255, 255),
+    emptyColor = Color.fromRgba(250, 250, 250, 255),
```
abigailalexander, line 48: The alpha values here should be 1, rather than 255. Phoebus uses 0-255 for transparency but CSS uses 0-1. Also, the default empty colour should be (192, 192, 192) to match Phoebus.

src/ui/widgets/EmbeddedDisplay/opiParser.ts
```diff
@@ -636,6 +637,7 @@ export const OPI_SIMPLE_PARSERS: ParserDict = {
   textAlignV: ["vertical_alignment", opiParseVerticalAlignment],
   backgroundColor: ["background_color", opiParseColor],
   foregroundColor: ["foreground_color", opiParseColor],
+  emptyColor: ["empty_color", opiParseColor],
```
abigailalexander, line 640: This property is called `empty_color` in bob files, but `color_fillbackground` in opi files. I'd suggest maybe moving this line into the `BOB_SIMPLE_PARSERS` and assigning `color_fillbackground` to `emptyColor` here. This way we can set the empty colour for both bob and opi (at the moment we can only set it for bob files this way).

abigailalexander (changes requested):
Looks great overall, I only have one major change. The bar chart doesn't fill the space available to the widget and looks quite thin (see pictures below). This is quite a big size difference between Phoebus and cs-web-lib, so it would be good to remove the border if possible?
Screenshot from 2025-09-18 09-48-31
Screenshot from 2025-09-18 09-48-18
Those side changes are all useful as well, thanks!

src/ui/widgets/Tank/tank.tsx
GregJHarris: I've made the correction

GregJHarris:
Reduced the border around the bar and scale.
Corrected the colors.
Applied a different mapping for emptyColor for bob and opi

PR #124, opened 2025-09-19
Add thermometer widget

### LINKED ISSUE #121: Add a Thermometer widget

GregJHarris:
Add a thermometer widget.
- This is drawn as an SVG using d3.

Main question is do we need an outline on the containing div? We have this on tank and progress bar.

Here are a few examples from the new widget:
image

...and the same bob file rendered in Pheobus:

image

closes #121


abigailalexander (changes requested):
Looks great overall, thanks a lot! Regarding the border on the containing div, I do think it would look better without it.

PR #126, opened 2025-09-25
Add meter widget

GregJHarris:
Add a meter widget, this uses react-gauge-component.
- We use a semi-circular gauge for all width / height ratios.

Some examples:
image

and corresponding Phoebus screens:
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -49,6 +49,7 @@
         "jsdom": "^25.0.1",
         "loglevel": "^1.9.2",
         "prettier": "^3.3.3",
+        "react-gauge-component": "^1.2.64",
         "react-id-generator": "^3.0.2",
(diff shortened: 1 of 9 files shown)


abigailalexander (changes requested):
Looks great overall, the scales seem good. I agree with not stretching the meter to odd proportions and maintaining the width/height ratio. One small fix - it seems like `scaledWidth` isn't being properly applied sometimes and the widget isn't taking up the full space available (nor being centered in the available space). I've attached a picture to show the difference between Phoebus and cs-web-lib for a widget with dimensions w: 440, h:370
Screenshot from 2025-09-26 10-23-52
Screenshot from 2025-09-26 10-28-43
As a result of the widget being smaller than expected, the scale is quite bunched up.

PR #127, opened 2025-10-08
Add Stripchart widget

abigailalexander:
This PR adds a StripChart widget using MUI X Charts, along with tests. Currently it only supports 1 PV, using the `yPv` value from the first trace. This is because we do not have multiple PV per widget capabilities yet.

Updating the `trace` and `axes` properties for StripChart affected the XYPlot widget, which was creating using .opi properties that are quite different. I've tried to reconcile the two different sets of properties as best as possible, but the XYPlot widget needs updating to .bob properties at some point in future for for functionality to be restored. This will be done in a separate PR (and could also remove the Plotly dependency to use MUI charts)

There are some Stripchart features not included e.g. `traceType = errorBars `and `showToolbar`. This is because these features seemed to require customisation on top of MUI which would have involved a lot more work than intended for a basic version of this widget.


GregJHarris (commented):
When using a black background the legend text, y-axis ticks and y-axis line are not visible.
image

GregJHarris (approved):
Thanks for the updates. Approved

PR #130, opened 2025-10-23
Add databrowser widget

abigailalexander:
This PR adds a Databrowser widget, which builds on top of the existing StripChart widget and adds archived data.
Other changes made:

- Added .plt file parser. Many of the properties are different to .bob and .opi and required their own parser
- In order to load the .plt file, many of the parsing functions have been made async to account for the file `fetch`. The Device and TabContainer components were also impacted by this and have been updated
- Modified some Striptool functionality - data is now stored in a `useRef` to prevent archived data being overwritten, also implemented `bufferSize` and `updatePeriod` properties to limit size of datasets and how often data updates.


GregJHarris (approved):
Approved Thank you

PR #131, opened 2025-11-06
Fix miscellaneous bugs

abigailalexander:
I've noticed several minor bugs recently, and have fixed them in this PR:

- The timestamp for PV updates is always in BST, even when we are in GMT. This was noticed recently when attempting to plot Databrowser data. The latest data was always 1 hour ahead of the current time. This seems to be a quirk of how the time was being set.
- Theming was not applied to DynamicPage child screens. Previously the MUI theme was used by default, now the Phoebus theme or another custom theme (if passed in) is applied. This can be seen on the B23 pmacController screen in Daedalus
- Default display background colour now matches Phoebus default
- Text centering for Readback widgets

### DIFF
--- a/src/connection/pvws.ts
+++ b/src/connection/pvws.ts
@@ -107,11 +107,8 @@ function pvwsToDType(data: any): DType {
   }
 
   let dtime = undefined;
-  if (data.seconds) {
-    const datetime = new Date(0);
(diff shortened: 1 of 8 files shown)


GregJHarris (approved):
Looks good to me

PR #176, opened 2026-03-05
Fix Arc widget

abigailalexander:
- Fixes the shape of the arc widget
- Updates tests
- Adjusts the radii of the arc widget to account for line width, which ensures the arcs are properly positioned


GregJHarris (approved):
Tested and looks good to me.

PR #185, opened 2026-04-08
Add missing DType export

abigailalexander:
Adds DType to the list of exported types, allowing it to be used in client applications.


GregJHarris (approved):
Approved, thanks for fixing

PR #190, opened 2026-04-29
Display grid layout widget

GregJHarris:
Adds a widget that uses the ReactGridLayout component, this component is similar to the Responsive component, but has only one layout and no breakpoints.


adedamola-sode (approved):
Tested and works as described

PR #191, opened 2026-04-30
Fix a bug that causes an exception when the default size is missing.

GregJHarris:
Fix a bug that causes an exception when the default size is missing.


abigailalexander (approved):
Looks good, thanks

PR #192, opened 2026-05-06
Queue script execution requests

GregJHarris:
This resolves the bug "When the synoptic is first loaded, pv dependent objects such as beamlines do not obey their scripts until the page is reloaded" https://jira.diamond.ac.uk/browse/EU-304

Now while the IFrame is starting up, scripts are put on a queue for execution once the IFrame is ready. This ensures that the scripts triggered at start up are executed.


abigailalexander (approved):
Looks good to me, thanks!

PR #193, opened 2026-05-11
Support for the display of an mjpg image stream

GregJHarris:
This PR:
- Allows the DemoImage widget to use a mjpg image stream by passing in the urls of the servers to the widget via the props.
- A default mjpg endpoint is defined in the configuration and used in the DynamicPage widget
- A primary mjpg endpoint can be passed to the DynamicPage widget


abigailalexander (approved):
Looks good to me, thanks

PR #194, opened 2026-05-12
Fix sizing when there are many absolute positioned elements

GregJHarris:
- Replaced the `<Box id={containerId} sx={{ width: "100%", height: "100%" }}>` container in the widget component with a fragment `<>`. The box was causing multiple relative divs to be placed in the layout - causing lots of empty scroll space.
- Removed the `<div style={fullStyle}>` container from DynamicPage, which now allows embedded display styles to fill the page.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@diamondlightsource/cs-web-lib",
-  "version": "0.10.12",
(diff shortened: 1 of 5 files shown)


abigailalexander (approved):
Tested and looks great, thanks!

PR #195, opened 2026-05-13
Parse Image props

abigailalexander:
- Adds parsing for Image widget props, such as color bar and regions of interest
- Color map tries to match names of colormaps to those available in h5web/lib, which Davidia uses 
- In the Image widget, I added prop defaults but left them commented out as they aren't used yet by the widget
- Reconfigured Axis prop to not use classes (this fixes a console error related to redux store serialisation)

### DIFF
--- a/src/ui/widgets/EmbeddedDisplay/bobParser.ts
+++ b/src/ui/widgets/EmbeddedDisplay/bobParser.ts
@@ -596,6 +596,8 @@ export const BOB_SIMPLE_PARSERS: ParserDict = {
   displayHorizontal: ["displayHorizontal", opiParseBoolean],
   xPv: ["xPv", opiParseString],
   yPv: ["yPv", opiParseString],
+  heightPv: ["heightPv", opiParseString],
+  widthPv: ["widthPv", opiParseString],
   axis: ["axis", bobParseNumber],
   pointType: ["point_type", bobParseNumber],
   pointStyle: ["point_style", bobParseNumber],
--- a/src/ui/widgets/EmbeddedDisplay/bobParser.ts
+++ b/src/ui/widgets/EmbeddedDisplay/bobParser.ts
@@ -41,7 +41,7 @@ import {
   BorderStyle,
   newBorder
 } from "../../../types/border";
-import { ColorUtils } from "../../../types/color";
+import { ColorBar, ColorUtils, newColorBar } from "../../../types/color";
 import { WidgetDescription } from "../createComponent";
 import { newPoint, newPoints, Point, Points } from "../../../types/points";
 import { Axis } from "../../../types/axis";
@@ -111,6 +111,7 @@ export const WIDGET_DEFAULT_SIZES: { [key: string]: [number, number] } = {
   ellipse: [100, 50],
   embedded: [400, 300],
   group: [300, 200],
+  image: [400, 300],
   label: [100, 20],
   led: [20, 20],
   linearmeter: [120, 120],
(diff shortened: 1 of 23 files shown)


src/ui/widgets/EmbeddedDisplay/bobParser.ts
```diff
@@ -358,7 +360,50 @@ function bobParseYAxes(props: any): Axis[] {
...
+
+/**
+ * Parses a color map, used in Heatmap and Image widgets
+ * @param props
+ * @returns
+ */
+function bobParseColorMap(props: any): string {
+  // Can also specify custom color maps with sections
+  // but for now we'll use named ones
+  const name = opiParseString(props.color_map.name);
```
GregJHarris: As props is of any type here, it might be worth considering handling undefined cases. Or at lest use the ?. operator, otherwise `props.color_map.name` could throw,

src/ui/widgets/EmbeddedDisplay/bobParser.ts
```diff
@@ -358,7 +360,50 @@ function bobParseYAxes(props: any): Axis[] {
...
+  return name;
+}
+
+/**
+ * Parses props for the color bar attached to a color map
+ * @param props
+ */
+function bobParseColorBar(props: any): ColorBar {
+  const parsedProps = parseChildProps(props.color_bar, BOB_SIMPLE_PARSERS);
```
GregJHarris: As above:
props.color_bar could throw if props is undefined.

src/ui/widgets/EmbeddedDisplay/bobParser.ts
```diff
@@ -358,7 +360,50 @@ function bobParseYAxes(props: any): Axis[] {
...
+}
+
+/**
+ * Parses an array of regions of interest
+ * @param props regions of interest
+ */
+function bobParseRois(props: any): Rois {
+  const rois: Rois = [];
+  let parsedProps = {};
+  if (props.rois.roi)
```
GregJHarris: props is any `props.rois.roi` could throw use `props?.rois?.roi`.

src/ui/widgets/EmbeddedDisplay/bobParser.ts
```diff
@@ -358,7 +360,50 @@ function bobParseYAxes(props: any): Axis[] {
...
+    if (props.rois.roi.length > 1) {
+      // If only one region, we are passed an object instead
+      // of an array
+      props.rois.forEach((roi: any) => {
```
GregJHarris: Bit confused here. the if statement uses `props.rois.roi.length` which suggests that `props.rois.roi` is an array and `props.rois` is an object. But `props.rois.forEach(...` suggests that `props.rois` is the array.
I would also suggest using `.map`, which feels a bit cleaner:
```
rois = props.rois.map((roi: any) => newRoi(parseChildProps(roi, BOB_SIMPLE_PARSERS)));
 ```

**Review by GregJHarris — CHANGES_REQUESTED** [review]
Thanks for converting the axis class to a plain object, that is really helpful.

Please could we also have some unit tests for the new methods in bobParser.ts.
Thanks

File: src/ui/widgets/EmbeddedDisplay/bobParser.ts (hunk 4 above)
Comment (abigailalexander) [reply src/ui/widgets/EmbeddedDisplay/bobParser.ts]: Good spot, I was trying to parse this like the Axes prop is parsed, but that's done slightly differently. I've updated this to use map

File: src/ui/widgets/EmbeddedDisplay/bobParser.ts [hunk 5]
```diff
@@ -358,7 +360,51 @@ function bobParseYAxes(props: any): Axis[] {
...
+  return newColorBar(parsedProps);
+}
+
+/**
+ * Parses an array of regions of interest
+ * @param props regions of interest
+ */
+export function bobParseRois(jsonProp: ElementCompact): Rois {
+  let rois: Rois = [];
+  let parsedProps = {};
```
GregJHarris: I'd suggest moving the declaration of parsedProps into the else block.
It is redefined in the inner scope of your map function and reassigned in the else block.

GregJHarris (commented):
One minor comment

src/ui/widgets/EmbeddedDisplay/bobParser.ts
abigailalexander: Thanks for the suggestion, the redefining bit is definitely a typo/bug XD I've implemented this change now

GregJHarris (approved):
Thanks for this work

PR #196, opened 2026-05-22
implement slide button

tomkane-dls:
I followed conventions from action and boolean buttons as much as possible

Default rendering from PVI screens:

Screenshot from 2026-05-22 12-41-02

Screenshot from 2026-05-22 12-41-24
Screenshot from 2026-05-22 12-41-32

### DIFF
--- a/src/phoebusTheme.ts
+++ b/src/phoebusTheme.ts
@@ -19,6 +19,12 @@ export const phoebusTheme = createTheme({
       onColor: "rgba(0,255,0,1)",
       offColor: "rgba(0,100,0,1)"
     },
+    slidebutton: {
+      main: "rgba(0,0,0,0)",
+      contrastText: "#000000",
+      onColor: "rgba(0,255,0,1)",
+      offColor: "rgba(0,100,0,1)"
+    },
     bytemonitor: {
       main: "#ffffffff",
       contrastText: "#000000",
--- a/src/phoebusTheme.ts
+++ b/src/phoebusTheme.ts
@@ -23,7 +23,7 @@ export const phoebusTheme = createTheme({
       main: "rgba(0,0,0,0)",
       contrastText: "#000000",
       onColor: "rgba(0,255,0,1)",
-      offColor: "rgba(0,100,0,1)"
+      offColor: "#D2D2D2"
     },
     bytemonitor: {
       main: "#ffffffff",
--- a/src/ui/widgets/SlideButton/slideButton.tsx
+++ b/src/ui/widgets/SlideButton/slideButton.tsx
@@ -1,0 +1,146 @@
+import React from "react";
+import { Widget } from "../widget";
+import { PVComponent, PVWidgetPropType } from "../widgetProps";
+import { registerWidget } from "../register";
+import {
+  InferWidgetProps,
+  ColorPropOpt,
+  IntPropOpt,
+  BoolPropOpt,
+  StringPropOpt,
+  FontPropOpt,
+  StringOrNumPropOpt
+} from "../propTypes";
+import { writePv } from "../../hooks/useSubscription";
+import { dTypeGetDoubleValue, newDType } from "../../../types/dtypes";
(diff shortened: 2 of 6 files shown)


src/phoebusTheme.ts
```diff
@@ -19,6 +19,12 @@ export const phoebusTheme = createTheme({
       onColor: "rgba(0,255,0,1)",
       offColor: "rgba(0,100,0,1)"
     },
+    slidebutton: {
+      main: "rgba(0,0,0,0)",
+      contrastText: "#000000",
+      onColor: "rgba(0,255,0,1)",
+      offColor: "rgba(0,100,0,1)"
```
abigailalexander: Phoebus seems to have the default off colour set to grey, instead of dark green

src/ui/widgets/SlideButton/slideButton.tsx
```diff
@@ -0,0 +1,146 @@
...
+const widgetName = "slidebutton";
+
+const SlideButtonProps = {
+  pvName: StringPropOpt,
+  onState: IntPropOpt,
+  offState: IntPropOpt,
+  onColor: ColorPropOpt,
+  offColor: ColorPropOpt,
+  onLabel: StringPropOpt,
+  offLabel: StringPropOpt,
```
abigailalexander: There aren't separate `offLabel` and `onLabel` props, just a single `label` prop. Currently the labels don't show because they aren't being parsed correctly here.

src/ui/widgets/SlideButton/slideButton.tsx
```diff
@@ -0,0 +1,146 @@
...
+import { Switch as MuiSwitch } from "@mui/material";
+import { getPvValueAndName } from "../utils";
+import { useStyle } from "../../hooks/useStyle";
+
+const widgetName = "slidebutton";
+
+const SlideButtonProps = {
+  pvName: StringPropOpt,
+  onState: IntPropOpt,
+  offState: IntPropOpt,
```
abigailalexander: `onState`, `offState`, `labelsFromPv` and `backgroundColor` aren't props that exist on this widget. Also, it would be good to add the `visible` prop too

abigailalexander (changes requested):
Thanks for adding this widget in. Looks good overall, just a few comments on props

src/ui/widgets/SlideButton/slideButton.tsx
tomkane-dls: Ah okay sorry I was comapring with other buttons in cs-web-lib rather than whats defined in Phoebus, I see - I'll compare with ToggleButton now

src/ui/widgets/SlideButton/slideButton.tsx
tomkane-dls: Ok yes same as below, thanks

src/ui/widgets/SlideButton/slideButton.tsx
```diff
@@ -0,0 +1,127 @@
...
+      ...props,
+      customColors: { onColor: props?.onColor, offColor: props?.offColor }
+    },
+    widgetName
+  );
+
+  const {
+    pvData,
+    enabled = true,
+    visible = true,
```
abigailalexander, line 54: `label` has a default value of "Label", rather than being an empty string/undefined

abigailalexander (commented):
Thanks for fixing the props!
I think left-aligning is the best way as this matches what Phoebus does, and ensures everything remains in the same layout as when the .bob file was edited/created. 
Phoebus:
Screenshot from 2026-05-26 13-20-34
cs-web-lib with centre alignment:
Screenshot from 2026-05-26 13-20-29
cs-web-lib with left-alignment:
Screenshot from 2026-05-26 13-29-59
Also noticed here while testing that it wraps text and hides the overflow, so it seems Phoebus isn't too concerned with the text overrun.

abigailalexander (approved):
All looks great now, thanks!

PR #197, opened 2026-05-26
XY plot support for bob files

GregJHarris:
Add basic support for xy plot capability for bob files using MUI XCharts. This PR:
- Adds basic support for XY plotting
- Fixes the broken plots_xy screen.
- Drop dependency on plotly.js

Outstanding:
- Error bar support.
- Detailed mapping of Phoebus functionality to XY ploy.


abigailalexander (approved):
Tested and the Example display plot looks like it's all working fine, thanks for fixing.

PR #199, opened 2026-05-28
XY Chart axis and colors

GregJHarris:
- Improve axis scales, including autoscale
- Support multiple y axis
- Improve handling of colors, including background and axis colors.
- Properly handle the legend

image

### DIFF
--- a/src/ui/widgets/XYPlot/xyPlot.tsx
+++ b/src/ui/widgets/XYPlot/xyPlot.tsx
@@ -41,6 +41,8 @@ import {
 
 const widgetName = "xyplot";
 
+const traceTypesWithoutLines = [0, 3];
+
 const XYPlotProps = {
   traces: TracesPropOpt,
   axes: AxesProp,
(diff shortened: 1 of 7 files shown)


src/ui/widgets/XYPlot/xyPlot.tsx
```diff
@@ -114,66 +133,87 @@ export const XYPlotComponent = (props: XYPlotComponentProps): JSX.Element => {
...
+              >
+                <ChartsTooltip trigger="axis" />
+                <ChartsAxisHighlight x="line" />
+                <BarPlot />
+                <LinePlot
+                  slotProps={{
+                    line: ({ seriesId }) => {
+                      const trace = traces?.[Number(seriesId)];
+                      // this hides the line if no line should be visible
+                      if (trace?.traceType === 0) {
```
abigailalexander: When I look at the example screen from Phoebus, I still see a line where it should be just marks. I think this is because the line should be hidden for traceType = 0 and also for = 3 (error bars with no line)

abigailalexander (approved):
Thanks, looks good. Just one really minor tweak

src/ui/widgets/XYPlot/xyPlot.tsx
GregJHarris: Thanks, good catch. Will resolve that.

GregJHarris: Done.

PR #200, opened 2026-05-28
Update Trace and Plt classes

abigailalexander:
Updates the Trace and Plt classes to be interfaces instead, due to redux toolkit errors. This now brings them in line with most of the other cs-web-lib types.


GregJHarris (approved):
Thanks for making this boilerplate change. Looks good to me.

PR #201, opened 2026-05-29
Fix Webcam css

abigailalexander:
The webcam css file has been updated to only apply properties to the webcam widget, instead of globally onto all img elements. The previous behaviour caused a bug where random img elements could appear the wrong size if they didn't set their own properties.
I also updated the Image widget to reset these properties here, so they should appear unchanged from before.


GregJHarris (approved):
Looks good to me, thanks.

PR #202, opened 2026-05-29
Add capability to draw markers to the xychart

GregJHarris:
- Support the parsing of marker object in xycharts of bob files.
- Render the markers as vertical reference lines in the xychart.

Omitted the capability to interact with the reference line. Phoebus supports PV write by interactively moving the marker line.

Appearance in cs-web-lib
image

Appearance in Phoebus:
image

### DIFF
--- a/src/ui/widgets/EmbeddedDisplay/bobParser.ts
+++ b/src/ui/widgets/EmbeddedDisplay/bobParser.ts
@@ -58,6 +58,7 @@ import {
   bobParseResponsiveMargins
 } from "./BobParsers/responsiveLayoutBobParser";
 import { newRoi, Rois } from "../../../types/rois";
+import { Markers, newMarker } from "../../../types/markers";
 
 const BOB_WIDGET_MAPPING: { [key: string]: any } = {
   action_button: "actionbutton",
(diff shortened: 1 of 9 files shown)


src/ui/widgets/EmbeddedDisplay/bobParser.ts
```diff
@@ -331,6 +332,26 @@ function bobParseTraces(props: any): Trace[] {
...
+ * Parses props from an array of markers
+ * @param props list of props for this element
+ * @returns a array of marker objects
+ */
+function bobParseMarker(props: any): Markers {
+  let markers: Markers = [];
+  if (props) {
+    // If only one marker, we are passed an object instead
+    // of an array
+    if (props?.marker?.length > 1) {
```
abigailalexander: If there's only one marker, it doesn't seem to get parsed here when it should be. I don't see a vertical line on the plot, but for two or more markers it works

abigailalexander (commented):
Thanks for adding this, looks good for multiple markers but spotted a small issue with single marker plots

src/ui/widgets/EmbeddedDisplay/bobParser.ts
GregJHarris: Thanks, looks like the XML to JSON parser converts something of the form:
```
<marker>
  <marker></marker>
</marker>
```
To `{ marker: {} }` rather than `{ marker: {}[] }`

I've added an array check in the code to ensure consistency.

abigailalexander (approved):
All looks good, thanks for the change

PR #203, opened 2026-06-05
Automatic layout of a pixel placed screen for React-Grid-Layout

GregJHarris:
This adds the capability to automatically layout react grid layout displays using pixel placed displays.
At the heart of it is `calculateDefaultLayout` which takes a set of child items a screen width, columns and cell height and computes a first pass layout using the position and width of the child items. Child items are wrapped if the screen width is too narrow to accommodate them.  The RGL verticalCompactor is then used to layout and resolve collisons in the y (columns). In addition for responsive screens the horizontalCompactor is also used.

Once a layout has been computed it is dispatched to update the file state, which causes a re-render using the new layout as the updated props are passed down.

I've moved the FileCache into a separate slice to improve separation of concerns.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@diamondlightsource/cs-web-lib",
-  "version": "0.10.15",
+  "version": "0.10.16",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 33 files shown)


abigailalexander (approved):
Looks great, thanks for this change!

PR #205, opened 2026-06-11
Add classfile parsing

abigailalexander:
This adds the basics of classfile parsing:

- Parse .bcf file based on filepath passed in via config
- Adds entries to theme based off class and widget name
- Override colours on a widget with class colour, if widget has a class set. If not, fall back to widget props
- Add tests

Currently only colours in the classes are applied. Border and font will be included in the MUI theme later, and other properties handled in another way.

### DIFF
--- a/src/ui/hooks/useClassFile.tsx
+++ b/src/ui/hooks/useClassFile.tsx
@@ -12,7 +12,7 @@ const keyMap: Record<string, string> = {
   foregroundColor: "contrastText"
 };
 
-const CLASS_PROPS = [
+const CLASS_PROPS = new Set([
   "offColor",
   "onColor",
   "foregroundColor",
@@ -26,11 +26,10 @@ const CLASS_PROPS = [
   "selectedColor",
   "deselectedColor",
   "borderColor"
-];
+]);
 
 export function useClassFile(userTheme?: Theme): Theme {
   const classFile = useSelector(selectClassFile);
-  //let theme = userTheme ?? phoebusTheme;
   const [theme, setTheme] = useState<Theme>(userTheme ?? phoebusTheme);
 
   useEffect(() => {
@@ -46,7 +45,7 @@ export function useClassFile(userTheme?: Theme): Theme {
     if (classFile !== undefined) {
       fetchData();
     }
-  }, [classFile]);
+  }, [classFile, userTheme]);
 
   return theme;
 }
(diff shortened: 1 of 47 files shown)


src/ui/hooks/useClassFile.tsx
```diff
@@ -0,0 +1,87 @@
...
+  "fillColor",
+  "needleColor",
+  "selectedColor",
+  "deselectedColor",
+  "borderColor"
+];
+
+export function useClassFile(userTheme?: Theme): Theme {
+  const classFile = useSelector(selectClassFile);
+  //let theme = userTheme ?? phoebusTheme;
```
GregJHarris: Stray comment

src/ui/hooks/useClassFile.tsx
```diff
@@ -0,0 +1,87 @@
...
+  const [theme, setTheme] = useState<Theme>(userTheme ?? phoebusTheme);
```
GregJHarris, line 33: In the event that userTheme changes, this does not update. It may be that we don't expect it to change in which case this is fine. Otherwise, we should add an update in the effect below.

src/ui/hooks/useClassFile.tsx
```diff
@@ -0,0 +1,87 @@
...
+    const matches = Object.entries(child)
+      .filter(([key]) => CLASS_PROPS.includes(key))
+      .map(([key, value]) => ({ key, value }));
+
+    // Assign colors to palette
+    palette[paletteName] = {
+      // Put Phoebus theme defaults, overwrite with class props
+      ...phoebusTheme.palette[widgetType],
+      ...Object.fromEntries(
+        matches.map(({ key, value }) => [keyMap[key] ?? key, value.colorString])
```
GregJHarris: Not sure if it is worth it, but in the event that the color is null or undefined value.colorString will throw. Maybe we should add a guard.

src/ui/hooks/useClassFile.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { selectClassFile } from "../../redux/slices/configurationSlice";
+import { phoebusTheme } from "../../phoebusTheme";
+
+// Map widget props to MUI theme props
+const keyMap: Record<string, string> = {
+  backgroundColor: "main",
+  foregroundColor: "contrastText"
+};
+
+const CLASS_PROPS = [
```
GregJHarris: I don't think this will have a huge impact on performance, but just as a note it is better to use a set here, it is faster.

`const CLASS_PROP_SET = new Set(CLASS_PROPS);`

`CLASS_PROP_SET.has(key);`  is O(1)
`CLASS_PROPS.includes(key)` is O(n)

GregJHarris (approved):
Approved with a few comments that you may want to concider.

PR #206, opened 2026-06-16
Border font classfile support

abigailalexander:
Adds border and font into the MUI theme from a classfile

### DIFF
--- a/src/ui/hooks/useClassFile.tsx
+++ b/src/ui/hooks/useClassFile.tsx
@@ -5,6 +5,8 @@ import { fetchAndConvert } from "./useFile";
 import { WidgetDescription } from "../widgets/createComponent";
 import { selectClassFile } from "../../redux/slices/configurationSlice";
 import { phoebusTheme } from "../../phoebusTheme";
+import { borderToCss } from "../../types/border";
+import { fontToCss } from "../../types/font";
 
 // Map widget props to MUI theme props
 const keyMap: Record<string, string> = {
@@ -12,7 +14,7 @@ const keyMap: Record<string, string> = {
   foregroundColor: "contrastText"
 };
 
-const CLASS_PROPS = new Set([
+const CLASS_COLOR_PROPS = new Set([
   "offColor",
   "onColor",
   "foregroundColor",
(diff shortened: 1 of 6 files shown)


src/ui/hooks/useClassFile.tsx
```diff
@@ -50,39 +59,75 @@ export function useClassFile(userTheme?: Theme): Theme {
...
+            keyMap[key] ?? key,
+            value?.colorString ?? undefined
+          ])
+        )
+      };
+
+    // Only colors go in the theme palette
+    const fontMatches = Object.entries(child)
+      .filter(([key]) => CLASS_FONT_PROPS.has(key))
+      .map(([key, value]) => ({ key, value }));
```
GregJHarris: Do we need to call to `map`? I don't think it does anything, after `Object.entries` we are already in the form {key: value}[] .
Could just use:
```
const fontMatches = Object.entries(child)
  .filter(([key]) => CLASS_FONT_PROPS.has(key));
```

src/ui/hooks/useClassFile.tsx
```diff
@@ -50,39 +59,75 @@ export function useClassFile(userTheme?: Theme): Theme {
...
+
+    // Assign fonts to typography
+    if (fontMatches.length > 0)
+      typography[paletteName] = {
+        ...Object.fromEntries(
+          fontMatches.map(({ key, value }) => [
+            keyMap[key] ?? key,
+            fontToCss(value) ?? undefined
+          ])
+        )
```
GregJHarris: Do we need the spread operator here and should we filter out undefined fonts? Eg:
```
typography[paletteName] = 
        Object.fromEntries(
          fontMatches.map(({ key, value }) => [
            keyMap[key] ?? key,
            fontToCss(value) ?? undefined
          ])
          .filter(([, v]) => v !== undefined)
        )
```

src/ui/hooks/useClassFile.tsx
```diff
@@ -50,39 +59,75 @@ export function useClassFile(userTheme?: Theme): Theme {
...
-    };
+    if (colorMatches.length > 0)
+      palette[paletteName] = {
+        // Put Phoebus theme defaults, overwrite with class props
+        ...phoebusTheme.palette[widgetType],
+        ...Object.fromEntries(
+          colorMatches.map(({ key, value }) => [
+            keyMap[key] ?? key,
+            value?.colorString ?? undefined
+          ])
```
GregJHarris: Should we filter out undefined values here?

GregJHarris (approved):
Approved with comments

PR #207, opened 2026-06-17
Create a cache of embedded display instances

GregJHarris:
- Add a `displayInstanceCache` to the file slice to store current instances of the displays built from individual files.
- Ensure that layouts are properly updated in the `displayInstanceCache` as child components are moved around.
- Add `widgetIdsCallback` to dynamic page. This will allow parent components to observe the display id and so get the display state from the `displayInstanceCache` to save it.
- Completely rewrite the embedded display and dynamic page tests so that they are true unit tests.
- Replace `uuidv4()` with `crypto.randomUUID()`
- Resolve high level package vulnerabilities, upgrade vite.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@diamondlightsource/cs-web-lib",
-  "version": "0.10.16",
+  "version": "0.10.17",
   "lockfileVersion": 3,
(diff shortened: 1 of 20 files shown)


abigailalexander (approved):
Looks good, thanks

PR #209, opened 2026-06-22
Add Style slice

abigailalexander:
- Adds a Style slice to the redux store, which stores all non-MUI theme class properties
- Update useStyle to overwrite props with class props
- Update tests

### DIFF
--- a/src/redux/slices/styleSlice.ts
+++ b/src/redux/slices/styleSlice.ts
@@ -1,0 +1,33 @@
+import { createSelector, createSlice, PayloadAction } from "@reduxjs/toolkit";
+
+export interface StyleStack {
+  classes: { [className: string]: any };
+}
+
+export const initialStyleState: StyleStack = {
+  classes: {}
+};
+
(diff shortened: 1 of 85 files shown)


src/redux/slices/styleSlice.ts
```diff
@@ -0,0 +1,33 @@
...
+  },
+  selectors: {
+    selectStyle: state => state.classes
+  }
+});
+
+export const { addClassStyle } = styleSlice.actions;
+export default styleSlice.reducer;
+
+export const selectStyle = (state: any) => state.style.classes;
```
GregJHarris: Is this necessary, we already define a selectStyle selector in the createSlice function, we just need to export it.
```
export const { selectStyle } = styleSlice .selectors;
```

GregJHarris (approved):

src/redux/slices/styleSlice.ts
abigailalexander: Good spot, thanks

PR #211, opened 2026-06-22
Adjust svg viewbox sizing

abigailalexander:
Correctly sizes viewboxes for the polyline and polygon widgets, which fixes an issue seen in the Daedalus i19 screens caused by svg boxes covering up symbols.

Also corrected a small bug where the transform was set to "rotation" instead of "rotate"


GregJHarris (approved):
Approved

PR #212, opened 2026-06-23
Delay path and macro resolution

GregJHarris:
This change moves the file URL absolute path normalisation and macro resolution from the file parser, to the create display instance stage. This ensures that the cached parsed file records do not contain partially resolved macros.

I have also moved some react and MUI dependencies into peer dependencies.


abigailalexander (approved):
Looks good to me, thanks

PR #214, opened 2026-06-25
Fix Image not updating

abigailalexander:
This fixes the issue of the Image component screens not updating when switching between camera screens, by adding a unique key and forcing a remount of the component on every render


GregJHarris (approved):
Looks good to me

PR #217, opened 2026-07-03
modified meter component to take in limits prop

adedamola-sode:
This now takes in the prop showLimits and passes it to the meter utility build subarcs


abigailalexander (approved):
Looks all good to me, thanks!

PR #221, opened 2026-07-20
Nl/ensure visibility is applied

NatLeung96:
This PR aims to resolve [EU-183](https://jira.diamond.ac.uk/browse/EU-183).

I think I got them all.


abigailalexander (approved):
Works correctly on all widgets, thanks for adding

PR #222, opened 2026-07-23
Enable converting display types for Quick Screens

abigailalexander:
This adds the functionality to convert a loaded file from one type to another. This can mostly be used for loading .bob files as quick screen type displays. If a string containing the type to convert to is passed into a DynamicPage/EmbeddedDisplay, the file contents are modified to add quick screen properties.

Also added tests to support this. Can test using branch convert-quick-screens on Daedalus.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@diamondlightsource/cs-web-lib",
-  "version": "0.10.20",
+  "version": "0.10.21",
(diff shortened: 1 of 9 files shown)


niamhdougan (approved):
As discussed:
- we probably don't want quick screen conversion from the top synoptic view
- quick screen menu buttons are staggeredly stacked when browser expanded on ultrawide monitor
otherwise works well :)

PR #223, opened 2026-07-24
Added option for parser to parse rule expressions

adedamola-sode:
The Opacity rule was not working for one of the example images, due to the fact that the bob file xml tag contains expression and not value, and the rules parser was looking for value.


abigailalexander (approved):
Looks good to me, thanks for fixing

PR #227, opened 2026-07-31
Added functionality to control darkmode of widgets from daedalus

adedamola-sode:
Added a new redux slice for the theme, which holds the value of the class, and is changed based on external functionality to set the right theme. Modified usestyles to use the theme from the redux store.


abigailalexander (changes requested):
This looks good to me overall, but I think it would make the most sense to have the `themeSlice` reducers etc just be a part of the `styleSlice` rather than in their own separate slice, since they're both handling the style from classfiles. If you could update the styleslice test if needed as well that would be great. Thanks!

abigailalexander (approved):
Thanks for the changes, looks all good to me!

PR #228, opened 2026-08-03
Update display instance hook

abigailalexander:
Updates the display instance hook to add an action that can create a display instance based on file contents, rather than a file path. This can be used when loading Quick Screens in daedalus, as they store file contents in browser local storage.

Can test with PR load-existing-quick-screen on Daedalus.


adedamola-sode (approved):
I was able to save with the save button, and then load with the load button, with each saved screen showing up in the load screen search tab. Although I'm unable to discern if the screens change due to my inability to modify the screens. Also, the load screen doesn't close when I click the load button...but it might be useful to browse screens, so I don't know if it's intentional. The search functionality works.

PR #229, opened 2026-08-06
groupbox to change colours when in darkmode

adedamola-sode:
The groupbox component does not use the useStyles, hook, So I've had to make some additions to if the system is in darkmode to change selected colours.

### DIFF
--- a/src/ui/widgets/GroupBox/groupBox.tsx
+++ b/src/ui/widgets/GroupBox/groupBox.tsx
@@ -68,6 +68,7 @@ export const GroupBoxComponent = (
   const effectiveForeground = style.colors.color ?? foregroundColor.colorString;
   const effectiveBorderColor =
     style.customColors?.lineColor ?? lineColor.colorString;
+  const effectiveStyleOpt = style.styleOpt ?? styleOpt;
 
   const outerDivStyle: CSSProperties = {
     width: "100%",
@@ -90,7 +91,7 @@ export const GroupBoxComponent = (
     ...fontToCss(font)
   };
 
-  if (styleOpt === 0) {
+  if (effectiveStyleOpt === 0) {
     // Typical group box with label
     outerDivStyle.paddingRight = "10px";
     boxStyle.paddingLeft = "8px";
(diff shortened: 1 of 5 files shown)


abigailalexander (commented):
Thanks for spotting Group Box was missing the class theming! Looking at this change, why are we not just adding the useStyles hook into the widget like the others?

Also, cs-web-lib should have no knowledge of custom class names that we are setting ourselves. All classfile logic should come from the parsed classfile and widget class name and be handled in useStyles, and we shouldn't be manually setting any colour themes that aren't explicit in Phoebus

src/ui/widgets/GroupBox/groupBox.tsx
```diff
@@ -41,17 +44,30 @@ const GroupBoxProps = {
...
+    lineColor = ColorUtils.fromRgba(0, 0, 0)
+  } = newProps;
+
+  const effectiveBackground = transparent
+    ? "transparent"
+    : (style.colors.backgroundColor ?? backgroundColor.colorString);
+  const effectiveForeground = style.colors.color ?? foregroundColor.colorString;
+  const effectiveBorderColor =
+    style.customColors?.lineColor ?? lineColor.colorString;
+  const effectiveStyleOpt = style.styleOpt ?? styleOpt;
```
abigailalexander: This should already be handled inside useStyles, so we should be able to use `style.colors`, `style.styleOpt` etc directly

src/ui/widgets/GroupBox/groupBox.tsx
```diff
@@ -105,12 +121,12 @@ export const GroupBoxComponent = (
           {styleOpt === 1 ? (
```
abigailalexander, line 112: Replace `styleOpt` with `style.styleOpt`

src/ui/widgets/GroupBox/groupBox.tsx
abigailalexander, line 128: This `styleOpt` needs to point towards the new `style.StyleOpt` too

abigailalexander (changes requested):
Thanks, just a few more very minor changes

abigailalexander (approved):

PR #230, opened 2026-08-07
Prevent widget action when dragging in quick screens. 

adedamola-sode:
added overlay on widgets when dragging in quick screen to prevent widget action.


abigailalexander (approved):
Tested and this works, thanks for the change!

PR #231, opened 2026-08-13
Added script to generate version file from package.json

adedamola-sode:
I tried other methods of versioning and exporting that for Daedalus to use. This method requires having a prerollup script that does the creating of a versioning file, whose variables are then exported.

### DIFF
--- a/.github/workflows/npm-publish-package.yml
+++ b/.github/workflows/npm-publish-package.yml
@@ -34,6 +34,6 @@ jobs:
           node-version: 24
           registry-url: https://registry.npmjs.org
       - run: npm ci
-      - run: npm run rollup
       - run: npm run generate-version
+      - run: npm run rollup
       - run: npm publish --access public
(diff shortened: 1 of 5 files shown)


abigailalexander (approved):
This looks good to me and works locally, thanks! I tested with the Daedalus branch One thing to check: the CI which builds and releases to npm doesn't include the `npm run all-checks` step, so I suspect the released version might not include the version.ts file like we expect. Could you add a step to run `npm run generate-version` in the npm publish CI?

.github/workflows/npm-publish-package.yml
```diff
@@ -35,4 +35,5 @@ jobs:
           registry-url: https://registry.npmjs.org
       - run: npm ci
       - run: npm run rollup
+      - run: npm run generate-version
```
abigailalexander, line 37: I think `generate-version` needs to run before the rollup command, otherwise it won't be included. I also get an error locally trying to run `rollup` before `generate-version`:
```
> @diamondlightsource/cs-web-lib@0.10.23 rollup
> rollup -c


src/index.ts → dist, dist...
[!] RollupError: Could not resolve "./version" from "src/index.ts"
```

abigailalexander (approved):

PR #232, opened 2026-08-21
Add delete widget support

abigailalexander:
This adds the ability to delete widgets from a responsive display if the mode is set to `editable`. On hover over a widget in editable mode, a delete button appears which removes the widget by id.


abigailalexander:
Can be tested with add-deleting-widget-quick-screens branch on Daedalus

adedamola-sode (approved):

